PR #1, opened 2018-12-20
define Token toString

zeroFruit:


junbeomlee:
In golang, `toString()`  should be `String()` and you need to implement it with test cases!

zeroFruit:
fixed it!

PR #4, opened 2018-12-21
Parse

junbeomlee:
1. ast.go
2. Refactored lex.go and add parser.go

### DIFF
--- a/parse/lex.go
+++ b/parse/lex.go
@@ -16,11 +16,16 @@
 
 package parse
 
+// emitter is the interface to emit the token to the client(parser).
+type emitter interface {
+	emit(t Token)
+}
+
 type Lexer struct {
 	tokench chan Token
 }
 
-func New(input string) *Lexer {
+func NewLexer(input string) *Lexer {
 
 	l := &Lexer{
 		tokench: make(chan Token, 2),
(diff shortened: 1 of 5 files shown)


parse/lex.go
```diff
@@ -32,18 +36,22 @@ func New(input string) *Lexer {
...
 
 	close(l.tokench)
 }
 
-type stateFn func(state) (stateFn, state, Token)
+func (l *Lexer) emit(t Token) {
+	l.tokench <- t
+}
+
+type stateFn func(state, emitter) stateFn
```
zeroFruit: I think stateFn need detail comment(ex. working as functional programming way). because this may can be our key feature in our lexer.

parse/lex.go
```diff
@@ -16,11 +16,15 @@
 
 package parse
 
+type emitter interface {
+	emit(t Token)
+}
```
zeroFruit: What is your purpose to add emitter interface?

parse/lex.go
```diff
@@ -32,18 +36,22 @@ func New(input string) *Lexer {
...
-	state := newState(input)
-	var token Token
+	state := state{
+		input: input,
+	}
 
 	for stateFn := DefaultStateFn; stateFn != nil; {
-		stateFn, state, token = stateFn(state)
-		l.tokench <- token
+		stateFn = stateFn(state, l)
```
zeroFruit: I think stateFn should have address of `state` if our lexer is going to have `state` as local variable

PR #9, opened 2018-12-24
Implement state next function and add test cases

junbeomlee:


parse/lex.go
```diff
@@ -111,6 +118,23 @@ func (s *state) cut(t TokenType) Token {
 	return token
 }
 
+// next returns the next rune in the input.
+func (s *state) next() rune {
+	if int(s.pos) >= len(s.input) {
+		s.width = 0
```
hihiboss, line 124: Doesn't it need to initialize `s.pos`??

parse/lex.go
```diff
@@ -98,11 +100,16 @@ func (l *Lexer) NextToken() Token {
 // state has the input(codes) as a string and has the current position and the line.
 type state struct {
 	input string
-	start int
-	pos   int
+	start Pos
+	pos   Pos
```
hihiboss: I think that `pos` name would be more specific because type `Pos` might confuse a variable name `pos` with type `Pos`.

parse/lex.go
junbeomlee: int types are initialized to `zero`. So it is fine

parse/lex.go
junbeomlee: I changed pos to end!!

parse/lex.go
hihiboss: Oh, I understood.

hihiboss (approved):

parse/lex.go
```diff
@@ -98,19 +100,41 @@ func (l *Lexer) NextToken() Token {
...
+		return Eof
+	}
+
+	r, w := utf8.DecodeRuneInString(s.input[s.end:])
+	s.width = Pos(w)
+	s.end += s.width
+
+	if r == '\n' {
+		s.line++
+	}
```
zeroFruit, line 134: I think it would be better if there's test case **which test changing `width` and `line`** for given input

junbeomlee: I have added line related a test case.

PR #11, opened 2018-12-26
Add vm structure

AgwaB:
details:

Add vm structure


junbeomlee:
I think we should talk about the structure tomorrow at the meeting.

PR #21, opened 2018-12-27
Edited README.md for statements used in Koa.

hihiboss:
details:

Modified README.md file to statements used in Koa.

Issue #22, opened 2018-12-27
[ast] define ast node, statement, literal


zeroFruit:
define AST node, statement, expression, literal structure

PR #23, opened 2018-12-27
Define TokenType, Ast node, statements, literals

zeroFruit:
resolved: #21, #22

details:

define TokenType, AST node, statement, expression, literal

### DIFF
--- a/parse/token.go
+++ b/parse/token.go
@@ -38,6 +38,10 @@ const (
 	String // "hello world"
 	Bool   // true, false
 
+	IntType
+	StringType
+	BoolType
+
 	Assign   // =
 	Plus     // +
 	Minus    // -
--- a/parse/token.go
+++ b/parse/token.go
@@ -33,6 +33,36 @@ const (
 	Illegal TokenType = iota
 
 	// Identifiers + literals
-	Ident // add, foobar, x, y, ...
-	Int   // 1343456
+	Ident  // add, foobar, x, y, ...
+	Int    // 1343456
+	String // "hello world"
+	Bool   // true, false
+
+	Assign   // =
+	Plus     // +
+	Minus    // -
+	Bang     // !
+	Asterisk // *
+	Slash    // /
+	Mod      // %
+
+	LT     // <
+	GT     // >
+	LTE    // <=
+	GTE    // >=
+	EQ     // ==
+	NOT_EQ // !=
+
+	Comma // ,
+
+	Lparen // (
+	Rparen // )
+	Lbrace // {
+	Rbrace // }
+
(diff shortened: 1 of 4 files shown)


hihiboss:
There is an import cycle error. `ast` imported `parse` and `parse` imported `ast`.

zeroFruit:
I solved cycle dependency problem by moving `ast.go` to `parse` package

junbeomlee:
We should think about this unused `statementNode()` function and `Node interface{}`
And i don't think the token is necessary!! This struct is not used at all in the book also.
```
type ReturnStatement struct {
 	Token <- this token
 	ReturnValue Expression
 }
```
So let's talk about this.

junbeomlee (changes requested):
We should think about this unused `statementNode()` function and `Node interface{}`
And i don't think the token is necessary!! This struct is not used at all in the book also.
```
type ReturnStatement struct {
 	Token <- this token
 	ReturnValue Expression
 }
```
So let's talk about this.

zeroFruit:
I agree with removing `Token` so fixed still keep `Node` interface and deeply consider about method names. and named Statement node for `do()` because statements basically **do something**, and Expression node for `produce()` because expression basically **produce at least one value**. Referenced from here:

https://stackoverflow.com/a/4730559/6433772

hihiboss:
~If `ast.go` is moved to `parse` package, is a dependency occurs?
Because, compiler component also uses `ast.go`...
Let's talk about whether moving `ast.go` is fine or not.~

zeroFruit:
@hihiboss @junbeomlee refactored!

parse/token.go
```diff
@@ -33,6 +33,36 @@ const (
 	Illegal TokenType = iota
 
 	// Identifiers + literals
-	Ident // add, foobar, x, y, ...
-	Int   // 1343456
+	Ident  // add, foobar, x, y, ...
+	Int    // 1343456
+	String // "hello world"
+	Bool   // true, false
+
```
junbeomlee, line 40: Int    // 1343456
String // "hello world"
Bool   // true, false

Also we need int type tokenType

zeroFruit:
Separated Token type with AST type (operator type, datastructure type) then compiler now only needs to know about AST type not Token type!

And parser needs to map Token type with AST type when building AST. Now the dependency is:
```
lexer -> Token
parser -> Token, AST
compiler -> AST
```
We can now remove dependency between Token and AST

junbeomlee (approved):
LGTM

hihiboss:
It’s explicit!

hihiboss (approved):

PR #32, opened 2018-12-28
Add checksig sample code

junbeomlee:
Add `checksig sample code` in readme.md


hihiboss:
How about using `###` instead of `####`?
It looks like sub-item of `### Contribution`

hihiboss:
Sample code is LGTM

junbeomlee:
The code should be under etc.
I made a mistake. I will fix it soon.

PR #33, opened 2018-12-28
Add vm structure

AgwaB:
details:
I made opcode logic in vm.
1. how about meke this thing to  opcode package?
2. or how about make function and handle in switch statement like a 'compiler in go'?

- [x] Test case

### DIFF
--- a/vm/instructions.go
+++ b/vm/instructions.go
@@ -1,0 +1,46 @@
+/*
+ * Copyright 2018 De-labtory
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package vm
+
+type InstructionBuffer interface {
+	GetItems(size int) []Item
+	JumpTo(i int)
+}
+
+type Instructions struct {
+	Buffer []Item
+	ip     int
+}
+
+func NewInstructions(instructions []Item) Instructions {
+	return Instructions{
+		Buffer: instructions,
+		ip:     0,
+	}
+}
+
+func (inst *Instructions) GetItems(size int) []Item {
+	item := make([]Item, 0)
+	item = append(item, inst.Buffer[inst.ip+1:inst.ip+1+size]...)
+
+	inst.ip += size
+	return item
+}
+
+func (inst *Instructions) JumpTo(i int) {
+	inst.ip += i
+}
--- a/vm/stack.go
+++ b/vm/stack.go
@@ -1,0 +1,41 @@
+/*
+ * Copyright 2018 De-labtory
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package vm
+
+type Stack struct {
+	item []Item
+	sp   int // Always points to the next value. Top of stack is stack[sp-1]
+}
+
+func NewStack() *Stack {
+	return &Stack{
+		item: make([]Item, 0),
+		sp:   0,
+	}
+}
+
+func (s *Stack) Push(item Item) {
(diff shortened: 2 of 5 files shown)


vm/vm.go
```diff
@@ -0,0 +1,131 @@
...
+	0x99:         Push{},
+}
+
+type Item []byte
+
+type VM struct {
+	stack       Stack
+	Instruction []byte
+}
+
```
junbeomlee: Instruction -> Instructions

vm/vm.go
```diff
@@ -0,0 +1,131 @@
...
+import (
+	"github.com/DE-labtory/koa/opcode"
+	"github.com/pkg/errors"
+)
+
+var opcodes = map[uint8]Opcode{
+	Add{}.Code(): Add{},
+	0x99:         Push{},
+}
+
```
junbeomlee, line 28: var opcodes = map[opcode.Type]Opcode{
 	opcode.Add: Add{},
 	0x99:         Push{},
 }

vm/vm.go
```diff
@@ -0,0 +1,131 @@
...
+	stack       Stack
+	Instruction []byte
+}
+
+type Stack struct {
+	item []Item
+	sp   int // Always points to the next value. Top of stack is stack[sp-1]
+}
+
+func New(instruction []byte) *VM {
```
hihiboss: Also the parameter has to be changed to “instructions”

vm/instructions.go
```diff
@@ -0,0 +1,45 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package vm
+
+type InstructionBuffer interface {
+	GetItems(size int) []Item
+	Move(i int)
```
zeroFruit: I think `Move` can have better meaningful func name, and I think `Move` functionality can be done inside `GetItems`

vm/vm.go
```diff
@@ -0,0 +1,86 @@
...
+	"github.com/pkg/errors"
+)
+
+var opcodes = map[opcode.Type]Opcode{
+	opcode.Add:  Add{},
+	opcode.Push: Push{},
+}
+
+type Item uint32
+
```
zeroFruit, line 30: imo `Item` too comprehensive

vm/instructions.go
AgwaB: good

vm/vm.go
AgwaB: instruction has opcodes and datas. I think Item could represent both. But on second thoughts, uint32 will be better because the processing of opcodes and datas is separated. What do you think?

vm/stack.go
```diff
@@ -0,0 +1,41 @@
...
+
+package vm
+
+type Stack struct {
+	item []Item
+	sp   int // Always points to the next value. Top of stack is stack[sp-1]
+}
+
+func NewStack() Stack {
+	return Stack{
```
zeroFruit: I think constructor should return pointer

vm/vm.go
```diff
@@ -0,0 +1,85 @@
...
+func Execute(instructions []Item) (Stack, error) {
+	stack := NewStack()
```
zeroFruit, line 32: I think `Execute` input params `instructions []Item` should be discussed with Compiler part(@hihiboss). 

1. In the case Compiler compile AST with go routine then input parameter might be better with `InstructionBuffer` because whenever VM need next instruction, we could fetch instruction from `InstructionBuffer`

2. In the case VM execute instructions after Compiler done compilation then `instructions []Item` would be good. but in this case, we don't need `InstructionBuffer` because we already get all the instructions from compiler

vm/vm.go
zeroFruit: What about `Instruction`? 
```go
func Execute(instructions []Item) (Stack, error) {
```

```go
type InstructionBuffer interface {
 	GetItems(size int) []Item
 }
```
I think in both case `Instruction` might be better than `Item` imo..

uint32 is ok I think :)

vm/vm.go
AgwaB: I think second reason seems to be problem of naming. The reason why i implement `InstructionBuffer` is to prevent direct access to instructions.
so it seems to require additional functions such as moving the pointer for operations such as a jump.

vm/stack.go
AgwaB: I did not feel a need for a pointer. but it looks good.

zeroFruit: I mean because the Stack methods have pointer receiver. so I thought that  if constructor return pointer, the user of `Stack` might be more convenient

vm/vm.go
zeroFruit: If the purpose of `InstructionBuffer` is to prevent direct access to instructions, then how about create `InstructionBuffer` concrete implementations **outside `Execute` function** using `instructions` then inject just `InstructionBuffer` to `Execute` parameters??

AgwaB: Sounds good. but vm will check some validation in `Execute` so `Execute` needs `instructions`. validation checking function is added...! sorry :)

vm/instructions.go
```diff
@@ -0,0 +1,46 @@
...
+	JumpTo(i int)
+}
+
+type Instructions struct {
+	Buffer []Item
+	ip     int
+}
+
+func NewInstructions(instructions []Item) Instructions {
+	return Instructions{
```
zeroFruit, line 30: Oh. this also would be nice if return pointer

PR #41, opened 2018-12-31
Parser create literal, identifier function skel

zeroFruit:
details:

Parser String, Boolean, Integer, Identifier parse func skeleton


junbeomlee (approved):
LGTM

PR #49, opened 2018-12-31
Define TokenTypeMap

zeroFruit:
Defined `TokenTypeMap` which is mapping `TokenType` with its token string.

I think this helps debugging or error checking


junbeomlee (approved):
LGTM

PR #50, opened 2018-12-31
Abstracted parseExpression func

### LINKED ISSUE #42: [parser] implement parseExpression

zeroFruit:
resolved: #42

Abstract `parseExpression` func.

### DIFF
--- a/parse/parser.go
+++ b/parse/parser.go
@@ -144,8 +144,33 @@ func parseStatement(buf TokenBuffer) (ast.Statement, []error) {
 	}
 }
 
+// TODO: create test cases :-)
+// ParseExpression parse expression in two ways, first
+// by considering expression as prefix, next as infix
+func parseExpression(buf TokenBuffer, pre precedence) (ast.Expression, []error) {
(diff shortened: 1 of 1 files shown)


junbeomlee:
How about naming `parseExpAsPrefix` to `parsePrefixExp` ?

zeroFruit:
We will have `parsePrefixExpression` function for parsing **real prefix expression**

In this case, This `parseExpAsPrefix` intention was as follow: 
*This might be prefix expression, so let's take prefix parse function from map first, then if prefix parse function exist parse this expression with that function*

so the **pseudo code** of `parseExpAsPrefix` might be
```
func parseExpAsPrefix(tokenBuffer) Expression {
    // *prefix* is prefix parsing function
    prefix := prefixFnMap[tokenBuffer.Peek(CURRENT)]
    if prefix == nil {
        return nil
    }

    return prefix(tokenBuffer)
}
```

junbeomlee:
ok

junbeomlee (approved):

PR #51, opened 2019-01-01
Init vm package

junbeomlee:
Initialized the default skeleton of the vm package.
I will add test case after review.
@leesd556

### DIFF
--- a/vm/asm.go
+++ b/vm/asm.go
@@ -1,0 +1,56 @@
+/*
+ * Copyright 2018 De-labtory
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package vm
+
+// Converts rawByteCode to assembly code.
+func assemble(rawByteCode []byte) (*asm, error) {
+	analysis()
+	return &asm{}, nil
+}
+
+// Do some analysis step (calculating the cost of running the code)
+func analysis() {
+
+}
+
+// Assemble Reader read assembly codes and can jump to certain assembly code
+type asmReader interface {
+	next() hexer
+	jump(i uint64)
+}
+
+type hexer interface {
+	hex() []uint8
+}
+
+type asm struct {
+	code  []hexer
+	cost  uint64
+	index uint64
+}
+
+func (a *asm) next() hexer {
+	return nil
+}
+
+func (a *asm) jump(pc uint64) {
+
+}
+
+func (a *asm) print() {
+
+}
--- a/vm/stack.go
+++ b/vm/stack.go
@@ -1,0 +1,72 @@
+/*
+ * Copyright 2018 De-labtory
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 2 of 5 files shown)


vm/asm.go
```diff
@@ -0,0 +1,56 @@
...
+type assembler interface {
+	next() Hexer
+	jump(pc uint64)
+}
+
+type Hexer interface {
+	hex() []uint8
+}
+
+type asm struct {
```
AgwaB, line 40: What about `instructions` rather than `codes`? 
I think that instruction has opcode and operand, also asm represent assembly code rather than machine code.

vm/vm.go
```diff
@@ -0,0 +1,76 @@
...
+	"fmt"
+
+	"github.com/DE-labtory/koa/opcode"
+)
+
+// The Execute function assemble the rawByteCode into an assembly code,
+// which in turn executes the assembly logic.
+func Execute(rawByteCode []byte) (*stack, error) {
+
+	s := &stack{}
```
AgwaB: why didn't you use `newStack`?

vm/vm.go
```diff
@@ -0,0 +1,76 @@
...
+	asm, err := assemble(rawByteCode)
+	if err != nil {
+		return &stack{}, err
+	}
+
+	for h := asm.next(); h != nil; {
+		if len(h.hex()) != 1 {
```
AgwaB: I think validation checking is already done in 'analysis`!

vm/stack.go
```diff
@@ -0,0 +1,72 @@
...
+
+// TODO: implement me w/ test cases :-)
+func (s *stack) len() int {
+	return len(s.items)
+}
+
+// TODO: implement me w/ test cases :-)
+// Print dumps the content of the stack
+func (s *stack) print() {
+	fmt.Println("### stack ###")
```
zeroFruit, line 63: this func for debugging?

vm/asm.go
```diff
@@ -0,0 +1,56 @@
...
+
+}
+
+// Assembler has assembly codes and the current program counter(position)
+type assembler interface {
+	next() Hexer
+	jump(pc uint64)
+}
+
+type Hexer interface {
```
zeroFruit: What Hexer do?

vm/stack.go
junbeomlee: yes!

vm/asm.go
junbeomlee: An hexer is an interface that can represent both data and opcode in assembly code.

vm/vm.go
junbeomlee: I forgot to use it!

vm/asm.go
junbeomlee: The instruction is like `opcode + data`. But some are `opcodes` and the others are `data`.  
It would be nice to have a word that can express both opcode and data.
I will think about it more!

vm/vm.go
junbeomlee: Yes but I added it for more safety.

junbeomlee:
I refactored the code and change some names.

vm/asm.go
```diff
@@ -0,0 +1,56 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package vm
+
+// Converts rawByteCode to assembly code.
+func assemble(rawByteCode []byte) (*asm, error) {
+	analysis()
```
zeroFruit, line 21: why don't you just return `asmReader` instead of `*ast`?

AgwaB:
LGTM

vm/asm.go
junbeomlee: The recommendation for golang is to get the `interface` as `parameter` and `return` the `concrete implementation`. This is more flexible way.

zeroFruit: ok! then.. LGTM!!

zeroFruit (approved):

PR #59, opened 2019-01-01
Add stack's testcase

AgwaB:
details:
stack testcase

- [x] Test case

### DIFF
--- a/vm/stack_internal_test.go
+++ b/vm/stack_internal_test.go
@@ -16,24 +16,80 @@
 
 package vm
 
-import "testing"
+import (
+	"testing"
+)
 
 func TestStack_new(t *testing.T) {
+	stack := newStack()
+	if cap(stack.items) != stackMaxSize {
+		t.Fatalf("Stack initializing failed")
+	}
+}
 
+func TestStack_pushN(t *testing.T) {
+	stack := newStack()
+
+	stack.pushN(item(1), item(2), item(3))
+
+	if len(stack.items) != 3 {
+		t.Fatalf("PushN did not work properly")
+	}
 }
 
 func TestStack_push(t *testing.T) {
+	stack := newStack()
 
+	stack.push(item(1))
+	stack.push(item(2))
+	stack.push(item(3))
+	if len(stack.items) != 3 {
+		t.Fatalf("Push did not work properly")
+	}
 }
 
 func TestStack_pop(t *testing.T) {
+	stack := newStack()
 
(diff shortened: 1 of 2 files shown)


vm/stack_internal_test.go
```diff
@@ -16,24 +16,69 @@
...
-import "testing"
+import (
+	"testing"
+
+	"github.com/magiconair/properties/assert"
+)
 
 func TestStack_new(t *testing.T) {
+	stack := newStack()
+	assert.Equal(t, stackMaxSize, cap(stack.items))
```
boohyunsik: Someone(including me) uses t.Fatalf method.
I think it is good to use one method to print error.
Or, is there any reason for using assert?

vm/stack_internal_test.go
```diff
@@ -16,24 +16,69 @@
 
 package vm
 
-import "testing"
+import (
+	"testing"
+
+	"github.com/magiconair/properties/assert"
```
zeroFruit: In the case we decided to use assert lib, then change with stretchr

vm/stack_internal_test.go
AgwaB: I used it for for no reason and @junbeomlee says we'll use built-in library as you said. 
so i'll change test methods

vm/stack_internal_test.go
```diff
@@ -16,24 +16,83 @@
...
+	stack := newStack()
+
+	stack.push(item(1))
+	stack.push(item(2))
+	stack.push(item(3))
 
+	if len(stack.items) != stack.len() {
+		t.Fatal("function len is invalid")
+	}
+	assert.Equal(t, len(stack.items), stack.len())
```
hihiboss: Is there a reason to compare sizes again?

vm/stack_internal_test.go
```diff
@@ -16,24 +16,83 @@
...
 }
 
 func TestStack_print(t *testing.T) {
```
hihiboss, line 81: I think this test is not necessary.

vm/stack_internal_test.go
AgwaB: i made a mistake... i fixed it!

vm/stack_internal_test.go
AgwaB: Just i wanted to check that method!

hihiboss:
goimport error!

junbeomlee (approved):

PR #70, opened 2019-01-02
Remove makeMockTokenBuffer and change mock token buffer as private

### LINKED ISSUE #68: [parser] refactoring parser test cases
related with #66

zeroFruit:
resolved: #68 

Actually there's not that much about refactoring just removed `makeMockTokenBuffer` and changed it into private.

Each test case has more lines and a little bit unreadable but I think this is worth for the future test cases

### DIFF
--- a/parse/parser_internal_test.go
+++ b/parse/parser_internal_test.go
@@ -21,60 +21,30 @@ import (
 	"testing"
 )
 
-type MockTokenBuffer struct {
+type mockTokenBuffer struct {
 	buf []Token
 	sp  int
 }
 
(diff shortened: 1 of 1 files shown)


boohyunsik:
Right. Firstly I made mocking function, but after adding test cases, I fill we did not need it.
Because we need more test cases.

zeroFruit:
Yes and per unit test case, I think we may have to provide different token set to `mockTokenBuffer` for each test case has different purpose

hihiboss:
@boohyunsik @zeroFruit 
I agree it.

hihiboss (approved):

Issue #13, opened 2018-12-27
[lex] Implement IdentifierStateFn


junbeomlee:
**Subject**
lex

**Detail**
We need to Implement IdentifierStateFn.
This function emit the identifier token

PR #71, opened 2019-01-02
Implement identifierStateFn, numberStateFn, spaceStateFn and helper f…

### LINKED ISSUE #12: [lex] Implement NumberStateFn
**Subject**
lexer

**Detail**
Implement `NumberStateFn`

joomanzi:
resolved: #12, #13

details:
1. Implement numberStateFn, spaceStateFn, identifierStateFn w/ testcase
2. Add isAlphaNumeric, isSpace function

- [ ] Test case

### DIFF
--- a/parse/lex.go
+++ b/parse/lex.go
@@ -17,6 +17,8 @@
 package parse
 
 import (
+	"strings"
+	"unicode"
 	"unicode/utf8"
 )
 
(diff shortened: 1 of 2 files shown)


parse/lex.go
```diff
@@ -249,23 +251,46 @@ func defaultStateFn(s *state, e emitter) stateFn {
...
+	//accept first character
+	s.next()
+
+	digits := "0123456789"
+	for strings.ContainsRune(digits, s.peek()) {
+		s.next()
+	}
 
+	e.emit(s.cut(Int))
 	return defaultStateFn
```
boohyunsik, line 276: Is there no change in defaultStateFn?

parse/lex_internal_test.go
```diff
@@ -243,3 +243,119 @@ func TestLex_defaultStateFn(t *testing.T) {
...
+		{'1'},
+		{'2'},
+		{'0'},
+		{'z'},
+		{'Z'},
+	}
+
+	for i, test := range tests {
+		if !isAlphaNumeric(test.input) {
+			t.Errorf("tests[%d] - %q is not alphanumeric", i, test.input)
```
boohyunsik, line 403: How about test the wrong case like &, *?
I think you can use assertPanic function.

parse/lex.go
```diff
@@ -249,23 +251,46 @@ func defaultStateFn(s *state, e emitter) stateFn {
...
 
-// TODO: implement me w/ test cases :-)
+// TODO: handle error case. panic isn't enough
+//
 // NumberStateFn scans an alphanumeric. ex) 123, 4001, 232
 // After reading Number, it returns DefaultStateFn.
+// number = { decimal_digit }
 func numberStateFn(s *state, e emitter) stateFn {
+	if r := s.peek(); !(r == '+' || r == '-' || ('0' <= r && r <= '9')) {
+		panic("Invalid Function call : numberStateFn")
```
zeroFruit: How about emit Illegal token (plz refer `token/token.go`)?? I think this is where Illigal token for.

If we panic program shut down and we can't find the reason of why the program shut down. plus if we emit illegal token instead of panic we can pass the responsibility to gracefully shutting down

+ oh I didn't see the todo...

parse/lex.go
```diff
@@ -249,23 +251,46 @@ func defaultStateFn(s *state, e emitter) stateFn {
...
 // IdentifierStateFn scans an identifiers. ex) a, b, add
 // After reading a identifier, it returns DefaultStateFn.
 //
 // when a input of a state is "int abc = 5" and start of state is 4,
 // IdentifierStateFn should emit "abc" and return DefaultStateFn
 //
+// identifier = letter { letter | unicode_digit }.
 func identifierStateFn(s *state, e emitter) stateFn {
+	if !(unicode.IsLetter(s.peek()) || s.peek() == '_') {
+		panic("Invalid Function call : identifierStateFn")
```
zeroFruit: same 👍

parse/lex.go
```diff
@@ -249,23 +251,46 @@ func defaultStateFn(s *state, e emitter) stateFn {
...
 }
 
-// TODO: implement me w/ test cases :-)
+// TODO: handle error case. panic isn't enough
+//
 // NumberStateFn scans an alphanumeric. ex) 123, 4001, 232
 // After reading Number, it returns DefaultStateFn.
+// number = { decimal_digit }
 func numberStateFn(s *state, e emitter) stateFn {
+	if r := s.peek(); !(r == '+' || r == '-' || ('0' <= r && r <= '9')) {
```
zeroFruit: this is optional, but we can refactor `!(r == '+' || r == '-' || ('0' <= r && r <= '9'))` in a similar way as `isSpace` or `isAlpha~` func

parse/lex_internal_test.go
```diff
@@ -243,3 +243,119 @@ func TestLex_defaultStateFn(t *testing.T) {
...
+		expectedType TokenType
+		expectedVal  string
+	}{
+		{"1", Int, "1"},
+		{"0", Int, "0"},
+		{"123", Int, "123"},
+		{"9990909", Int, "9990909"},
+		{"+909", Int, "+909"},
+		{"-909", Int, "-909"},
+	}
```
zeroFruit, line 263: Later we should test when illegal token comes to `numberStateFn`

parse/lex.go
joomanzi: thx for review!
Before fixing my code, I need some advice as beginner!

1) I leave that issue in TODO list because we don't have err message section in state struct.
even we shut down program gracefully via 'Illegal token',  how can i leave error message?

2) do we really need to check condition check? 
Apparently, doing double-check would become more safe. 
but I found out even Go SDK lex.go don't check twice time.

forgive my innocent lol

parse/lex.go
joomanzi: yes. can say only state might change ( e.g., start, end pos)

parse/lex.go
zeroFruit: 1. I think you can leave a message using `Token.Val`, because token type is illegal we can manage illegal reason with value property
2. don't understand 'check condition check', which code?

parse/lex_internal_test.go
junbeomlee: I agree!

parse/lex.go
joomanzi: confirmed! appreciate u  :D

parse/lex.go
joomanzi: i rather change my logic. it looks better to me! due to added testcase

parse/lex_internal_test.go
joomanzi: thx.  i fix my bug and add testcase

junbeomlee (approved):
LGTM

Issue #61, opened 2019-01-01
[vm] Implement opcode's instruction and testcase


AgwaB:
**Subject**
opcode.go, vm.go

**Detail**
In vm.go, opcode's instruction should be implemented.

ex) `Add`, `Push`, `Pop`, `Mul`, `Sub`, `Div`, `Mod` ...

kwongiho:
Assigned to me

AgwaB:
@KwonGiho ok! but after merged #72, i'll implement basic opcode `Add`, `Push` to show example!

Issue #60, opened 2019-01-01
[vm] Implement analysis function


AgwaB:
**Subject**
asm.go

**Detail**
we need to analyze while assemble the `rawByteCode`,
analysis() function calculate the cost of running the code.
And we need to discuss about it more.

ex) if code's cost is very, vm may possibly stop.

**Reference**
Ethereum opcode gas costs
https://docs.google.com/spreadsheets/d/1n6mRqkBz3iWcOlRem_mO09GtSKEKrAsfO7Frgx18pNU/edit#gid=0

PR #72, opened 2019-01-02
Implement assemble func prototype

AgwaB:
details:
We need to discuss about #60 
But when we implement opCodes( #61 ), it needs `asm.code` 
( it is analyzed `rawByteCode` but the analysis is focused on cost maybe. )

I think it is better to implement the analysis properly after it is fully discussed!! so i implement prototype of `analysis`...

- [x] Test case

### DIFF
--- a/vm/asm.go
+++ b/vm/asm.go
@@ -16,15 +16,47 @@
 
 package vm
 
+import (
+	"github.com/DE-labtory/koa/opcode"
+	"github.com/pkg/errors"
+)
+
+var ErrInvalidOpcode = errors.New("Invalid byteCode")
+
+var opCodes = map[opcode.Type]opCode{
+	opcode.Add:  add{},
+	opcode.Push: push{},
+}
+
 // Converts rawByteCode to assembly code.
 func assemble(rawByteCode []byte) (*asm, error) {
-	analysis()
-	return &asm{}, nil
+	asm := newAsm()
+
+	for i := 0; i < len(rawByteCode); i++ {
+		op, ok := opCodes[opcode.Type(rawByteCode[i])]
+
+		if !ok {
+			return nil, ErrInvalidOpcode
+		}
+
+		switch op.hex()[0] {
+		case uint8(opcode.Push):
+			body := make([]uint8, 0)
+			body = append(body, rawByteCode[i+1:i+5]...)
+
+			asm.code = append(asm.code, op)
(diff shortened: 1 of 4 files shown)


AgwaB:
Moved `analysis` logic to `assemble`

AgwaB:
I think `uint32To4Bytes` is not a good naming. Anyone have any other opinion?

vm/asm.go
```diff
@@ -16,15 +16,45 @@
...
-	analysis()
-	return &asm{}, nil
+	asm := newAsm()
+
+	for i := 0; i < len(rawByteCode); i++ {
+		op, ok := opCodes[opcode.Type(rawByteCode[i])]
+
+		if !ok {
+			return nil, errors.New("Invalid byteCode")
+		}
```
junbeomlee, line 40: I think it's better to predefine 
`var ErrInvalidOpcode = errors.New("Invalid byteCode")` and return `ErrInvalidOpcode`

vm/asm_internal_test.go
```diff
@@ -0,0 +1,124 @@
...
+	rawByteCode = append(rawByteCode, uint8(opcode.Push))
+	rawByteCode = append(rawByteCode, uint32To4Bytes(1)...)
+	rawByteCode = append(rawByteCode, uint8(255)) // invaild code
+	rawByteCode = append(rawByteCode, uint32To4Bytes(300)...)
+	rawByteCode = append(rawByteCode, uint8(opcode.Add))
+
+	_, err := assemble(rawByteCode)
+	if err == nil {
+		t.Error("The desired error was not found ")
+	}
```
junbeomlee, line 76: It would be better to compare like this
```
if err != ErrInvalidOpcode{
    t.Error("The desired error was not found ")
}
```

vm/asm_internal_test.go
```diff
@@ -0,0 +1,124 @@
...
+	defer func() {
+		if r := recover(); r == nil {
+			t.Fatal("Access to in valid program counter!")
+		}
+	}()
+	asm.jump(1)
+	asm.jump(2)
+	asm.jump(4)
+	asm.jump(10) // Error occur
+}
```
junbeomlee, line 163: It is important to distinguish between cases where err occurs and cases that do not occur. 
In cases where err does not occur, it is necessary to check whether asm's pc value is correctly jumped.

junbeomlee (approved):
LGTM

PR #85, opened 2019-01-05
Encode fix

### LINKED ISSUE #82: [encode] Fix int64 type
**Where**
encode

**Bug detail**
We use int32 not int64.

hihiboss:
resolved: #82 

details:

Fixed encoding.

- [x] Test case

### DIFF
--- a/encoding/encode.go
+++ b/encoding/encode.go
@@ -46,17 +46,15 @@ func EncodeOperand(operand interface{}) ([]byte, error) {
 // Encode integer to hexadecimal bytes
 // ex) int 123 => 0x7b
 func encodeInt(operand int) ([]byte, error) {
-	operand32 := int32(operand)
-	s := fmt.Sprintf("%x", operand32)
-
-	// Encoded byte length should be even number
-	if len(s)%2 == 1 {
-		s = "0" + s
-	}
+	s := convertTo4Bytes(operand)
 
 	b, err := hex.DecodeString(s)
 	if err != nil {
 		return nil, err
+	}
+
+	if len(b) != 4 {
+		return nil, errors.New("Integer size is not 32 bits")
 	}
 
 	return b, nil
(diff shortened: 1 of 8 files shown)


zeroFruit:
I think it is duplicate with #86 ?

hihiboss:
@zeroFruit 

Because #86 came out from this, it looks like this is duplicated!

encoding/encode.go
```diff
@@ -18,36 +18,49 @@ package encoding
...
 	switch op := operand.(type) {
 	case int:
-		return nil
+		i, err := encodeInt(op)
+		if err != nil {
+			log.Fatal(err)
+		}
+
+		return i
+
```
junbeomlee, line 34: It would be better to hand it over to return rather than to handle the error here.
```
func EncodeOperand(operand interface{}) ([]byte,error){
    return encodeInt(op)
}
```

encoding/encode.go
```diff
@@ -18,36 +18,49 @@ package encoding
...
 	}
 }
 
-// Encode integer to hexadecimal bytes
-// ex) int 123 => 0x7b
+// EncodeOperand integer to hexadecimal bytes
+// ex) int 123 => 7b
 func encodeInt(operand int) ([]byte, error) {
-	s := strconv.FormatInt(int64(operand), 16)
+	s := fmt.Sprintf("%x", operand)
```
junbeomlee: Returning size of []byte should be 32bits is it?

encoding/encode.go
hihiboss: I fixed!

encoding/encode.go
hihiboss: I edited it. Plz check and commit it!

junbeomlee:
Did u check the size of bytes?!

junbeomlee (approved):

junbeomlee:
Good work!

PR #86, opened 2019-01-05
Compiler skeleton

### LINKED ISSUE #87: [compiler] Define compiler skeleton
**Subject**
compiler

**Detail**
Define compiler skeleton.

hihiboss:
resolved: #87

details:

1. Defined compiler skeleton.
2. Implemented `compileBoolean()` function.
3. Implemented `emit()` function.
4. Added `output` struct. This has `bytecode`, `asmcode`, `pc`. If you have better name than `output`, please comments it!

Important functions are `Compile()`, `compileNode()`, `emit()`.

- [x] Test case


translate/compiler.go
```diff
@@ -16,15 +16,130 @@
...
+	default:
+		return errors.New("compileNode() error - " + node.String() + " could not compiled")
+	}
+
 	return nil
 }
 
-func (c *Compiler) Compile(program ast.Program) {
+// TODO: implement w/ test cases :-)
+func (c *Compiler) compileIdentifier(node ast.Identifier) error {
```
AgwaB: Are these functions ethereum style?

hihiboss: Yes they are, but I mixed the go ethereum style and the monkey style.


Go ethereum seperates only element, number, eof, etc. (go ether reads line by line.)
Monkey seperates string, integer, boolean, prefix, infix, etc. But monkey doesn’t make functions. It is only written by switch/case statements.

So, I seperates them like the monkey, but functions style came from go ethereum.

zeroFruit:
I think it is duplicate with #85 ?

zeroFruit:
this looks good to me

zeroFruit (dismissed):

junbeomlee (changes requested):
i have some comments!!

hihiboss:
@junbeomlee

I can't see your comments! How could I see yours??

boohyunsik:
Plz solve conflict :-)

hihiboss:
@boohyunsik

I solved it!

translate/asm.go
```diff
@@ -0,0 +1,59 @@
...
+		return "Pop"
+	case 0x21:
+		return "Push"
+	case 0x22:
+		return "Mload"
+	case 0x23:
+		return "Mstore"
+
+	default:
+		return ""
```
boohyunsik: No error?

junbeomlee:
i think the output struct should be exposed structure!! so plz rename output to Output and i agree that there will be a better name than output.

translate/asm.go
```diff
@@ -0,0 +1,59 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package translate
+
+import "github.com/DE-labtory/koa/opcode"
+
+type output struct {
```
AgwaB: 1. What is the difference between `bytecode` and `asmcode`? I think it's meant to show user-friendly type (`string`) am i right?

2. What is the purpose of `pc`?

translate/compiler.go
```diff
@@ -16,15 +16,124 @@
...
-type Compiler struct {
+	"github.com/DE-labtory/koa/ast"
+	"github.com/DE-labtory/koa/encoding"
+	"github.com/DE-labtory/koa/opcode"
+)
+
+// Compile() compiles statements in ast.program.
+// Statements would be compiled to byte code.
+// TODO: implement w/ test cases :-)
+func Compile(program ast.Program) ([]byte, error) {
```
AgwaB: Why does not this return an output, but return a [] byte?

translate/asm.go
```diff
@@ -0,0 +1,59 @@
...
+	bytecode []byte
+	asmcode  []string
+	pc       uint64
+}
+
+func opcodeToString(p opcode.Type) string {
```
AgwaB: I think this seems to be in constant.go.

translate/asm.go
hihiboss: I'm thinking whether return `error` or `""`(empty string) and so on. What do you think?

translate/asm.go
hihiboss: 1. `bytecode` means real raw byte code which you already know. (ex 0x1234...) And `asmcode` means assemble code like `Push 0x01 Pop ...`. It is necessary because of `Jump`. It is also user-friendly type too.

2. `pc` would be used in calculating next branch. `Jump` changes execution flow, so it needs to calculate next or return branch position.

translate/compiler.go
hihiboss: Because `vm` only needs to know `[]byte` (bytecode), not assemble code. But I'm thinking about returning `output` or only `[]byte`.

translate/asm.go
hihiboss: Yes, you're right! I'll move it!

translate/compiler.go
AgwaB: Yes vm only needs to know []byte. But `asmcode` is also needed as a result of the compile

translate/asm.go
AgwaB: ah, I have now seen the `output` used in the compile process...!  good

translate/asm.go
boohyunsik: If an invalid input is received, program should stop. 
so I think we can handle error when opcode converting, or we can handle it after empty string received.
for example, 
```go
str := opcodeToString()
if str == nil {
   return nil, err
}
```
But, I think handling error where `the error occured` is better.

AgwaB: I think compiler don't need to check the error about `emit`. because it's internal process of koa  whereas `vm` is independent. 
and if we should do error checking, is it more go style to return (`string`, `err`)?
```go
str, err := opcodeToString()
if err != nil {
   return err
}
```

translate/compiler.go
hihiboss: Then, I'll return `output` not `[]byte`!

translate/asm.go
hihiboss: I'll modify this with go style! (`return (string, error)`)

PR #89, opened 2019-01-06
VM opcode skeleton

AgwaB:
details:

1. Renamed `assemble` -> `disassemble`
2. Defined opcodes skeleton
3. Implement negative numbers operate

ᅟI have a issue here.
Ethereum use `big.int` and use the operation method in `big.int` like a `big.Add`, `big.Sub`, `big.Mul`, `big.Div`, `big.Mod`. but `big.Div` and `big.Mod` use Euclidean division (unlike Go!!)

For example,  
int a = -20
int b = 30
result of a/b is -1 and a%b is 10. not a 0 and 10.

plz review...! :)

### DIFF
--- a/vm/asm.go
+++ b/vm/asm.go
@@ -22,6 +22,10 @@ import (
 
 var opCodes = map[opcode.Type]opCode{
 	opcode.Add:  add{},
+	opcode.Mul:  mul{},
+	opcode.Sub:  sub{},
+	opcode.Div:  div{},
+	opcode.Mod:  mod{},
 	opcode.Push: push{},
 }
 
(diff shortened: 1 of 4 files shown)


zeroFruit:
lgtm 👍

zeroFruit (approved):

hihiboss (approved):
LGTM!

AgwaB:
@junbeomlee

zeroFruit:
And it is just question. why we use `big` ?

AgwaB:
@zeroFruit `big` has well implemented library, especially `Div` and `Mod`. that's why i used `big`. 
But i will modify it to local method... because I think it is too big to use

junbeomlee:
I agree with @zeroFruit.
We can implement `div`, `mod` for 32bit!

junbeomlee:
Nice work!

PR #90, opened 2019-01-06
Add parser documentation to README.md

zeroFruit:
Add documentation about parser. and image is prototyped so I think this need @junbeomlee's help


junbeomlee:
looks good!!
But I'll redraw the picture.

junbeomlee (approved):
Nice works!

PR #94, opened 2019-01-07
Implement infix parser

### LINKED ISSUE #80: [parser] implement parseInfixExpression
```
// TODO: implement me w/ test cases :-)
func parseInfixExpression(buf TokenBuffer, exp ast.Expression) (ast.Expression, []error) {
	return nil, nil
}
```

boohyunsik:
resolved: #80 

1. Implement InfixExpression struct in ast.go
2. Implement Infix parse methods
3. Implement test cases about above ones

- [x] Test case

### DIFF
--- a/ast/ast.go
+++ b/ast/ast.go
@@ -156,3 +156,16 @@ func (pe *PrefixExpression) produce() {}
 func (pe *PrefixExpression) String() string {
 	return ""
 }
+
+// Represent infix expression
+type InfixExpression struct {
+	Operator
+	Left  Expression
+	Right Expression
+}
+
+func (ie *InfixExpression) produce() {}
+
+func (ie *InfixExpression) String() string {
+	return ""
+}
--- a/ast/ast.go
+++ b/ast/ast.go
@@ -16,6 +16,11 @@
 
 package ast
 
+import (
+	"fmt"
+	"strconv"
+)
+
 // Node represent ast node
 type Node interface {
 	String() string
@@ -127,7 +132,7 @@ type IntegerLiteral struct {
 func (il *IntegerLiteral) produce() {}
 
 func (il *IntegerLiteral) String() string {
-	return string(il.Value)
+	return strconv.FormatInt(il.Value, 10)
 }
 
 // Represent Boolean expression
@@ -157,15 +162,15 @@ func (pe *PrefixExpression) String() string {
 	return ""
 }
 
-// Represent infix expression
+// Repersent Infix expression
 type InfixExpression struct {
+	Left Expression
 	Operator
-	Left  Expression
 	Right Expression
 }
 
 func (ie *InfixExpression) produce() {}
 
 func (ie *InfixExpression) String() string {
-	return ""
+	return fmt.Sprintf("(%s %s %s)", ie.Left.String(), string(ie.Operator.Val), ie.Right.String())
 }
--- a/parse/parser.go
+++ b/parse/parser.go
@@ -61,6 +61,22 @@ var precedenceMap = map[TokenType]precedence{
 	Rbrace: LOWEST,
 }
 
+// Operator map
+var operatorTypeMap = map[TokenType]ast.OperatorType{
+	Plus:     ast.Plus,
+	Minus:    ast.Minus,
+	Bang:     ast.Bang,
+	Asterisk: ast.Asterisk,
+	Slash:    ast.Slash,
+	Mod:      ast.Mod,
+	LT:       ast.LT,
+	GT:       ast.GT,
+	LTE:      ast.LTE,
+	GTE:      ast.GTE,
+	EQ:       ast.EQ,
+	NOT_EQ:   ast.NOT_EQ,
+}
+
 // PeekNumber restrict peek count from the TokenBuffer
 type peekNumber int
 
@@ -104,11 +120,16 @@ func expectNext(buf TokenBuffer, t TokenType) (bool, error) {
 }
 
 func curPrecedence(buf TokenBuffer) precedence {
-	return precedenceMap[buf.Peek(CURRENT).Type]
+	if p, ok := precedenceMap[buf.Peek(CURRENT).Type]; ok {
+		return p
+	}
+	return LOWEST
 }
-
 func nextPrecedence(buf TokenBuffer) precedence {
-	return precedenceMap[buf.Peek(NEXT).Type]
+	if p, ok := precedenceMap[buf.Peek(NEXT).Type]; ok {
+		return p
+	}
+	return LOWEST
 }
 
 // ParseError contains error which happened during
@@ -157,7 +178,6 @@ func parseStatement(buf TokenBuffer) (ast.Statement, []error) {
 	}
 }
 
-// TODO: create test cases :-)
 // ParseExpression parse expression in two ways, first
 // by considering expression as prefix, next as infix
 func parseExpression(buf TokenBuffer, pre precedence) (ast.Expression, []error) {
@@ -200,26 +220,56 @@ func makePrefixExpression(buf TokenBuffer) (ast.Expression, []error) {
 	return exp, errs
 }
 
-// TODO: implement me w/ test cases :-)
 // MakeInfixExpression retrieves infix parse function from map
 // then parse expression with that function if exist.
 func makeInfixExpression(buf TokenBuffer, exp ast.Expression, pre precedence) (ast.Expression, []error) {
-	return nil, nil
+	errs := make([]error, 0)
+	expression := exp
+
+	for pre < curPrecedence(buf) {
+		fn := infixParseFnMap[buf.Peek(CURRENT).Type]
+		if fn == nil {
+			errs = append(errs, parseError{buf.Peek(CURRENT).Type, "infix parse function not defined"})
+			return nil, errs
+		}
+
+		var err []error
+		expression, err = fn(buf, expression)
+		if len(err) > 0 {
+			return nil, append(errs, err...)
+		}
+	}
+	return expression, nil
 }
 
-// TODO: implement me w/ test cases :-)
-func parseInfixExpression(buf TokenBuffer, exp ast.Expression) (ast.Expression, []error) {
-	return nil, nil
+func parseInfixExpression(buf TokenBuffer, left ast.Expression) (ast.Expression, []error) {
+	curTok := buf.Read()
+	expression := &ast.InfixExpression{
+		Left: left,
+		Operator: ast.Operator{
+			Type: operatorTypeMap[curTok.Type],
+			Val:  ast.OperatorVal(curTok.Val),
+		},
+	}
+	precedence := precedenceMap[curTok.Type]
+
+	var err []error
+	expression.Right, err = parseExpression(buf, precedence)
+	if len(err) > 0 {
+		return nil, err
+	}
+
+	return expression, nil
 }
 
 func parseIdentifier(buf TokenBuffer) (ast.Expression, []error) {
 	errs := make([]error, 0)
-	token := buf.Peek(CURRENT)
+	token := buf.Read()
+
 	if token.Type != Ident {
 		errs = append(errs, errors.New("parseIdentifier() - "+token.Val+" is not a identifier"))
 		return nil, errs
 	}
-	buf.Read()
 	return &ast.Identifier{Value: token.Val}, nil
 }
 
--- a/parse/parser_internal_test.go
+++ b/parse/parser_internal_test.go
@@ -40,7 +40,7 @@ func (m *mockTokenBuffer) Peek(n peekNumber) Token {
 
 var mockError = errors.New("error occurred for some reason")
 
-func TestParse_curTokenIs(t *testing.T) {
+func TestCurTokenIs(t *testing.T) {
 	tokens := []Token{
 		{Type: Int, Val: "1"},
 		{Type: Ident, Val: "ADD"},
@@ -84,7 +84,7 @@ func TestParse_curTokenIs(t *testing.T) {
 	}
 }
 
-func TestParse_nextTokenIs(t *testing.T) {
+func TestNextTokenIs(t *testing.T) {
 	tokens := []Token{
 		{Type: Int, Val: "1"},
 		{Type: Ident, Val: "ADD"},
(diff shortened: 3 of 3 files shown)


parse/parser.go
```diff
@@ -200,16 +202,44 @@ func makePrefixExpression(buf TokenBuffer) (ast.Expression, []error) {
...
+			return nil, append(errs, err...)
+		}
+	}
+	return expression, nil
 }
 
-// TODO: implement me w/ test cases :-)
 func parseInfixExpression(buf TokenBuffer, exp ast.Expression) (ast.Expression, []error) {
-	return nil, nil
+	expression := &ast.InfixExpression{Left: exp,
```
zeroFruit: I think there's no test case for this function

zeroFruit: When we are executing this function, we can fully assume that the current token is **already** infix operator 

so before making `expression` we can `buf.Read()` and retrieve token.

parse/parser.go
```diff
@@ -200,16 +202,44 @@ func makePrefixExpression(buf TokenBuffer) (ast.Expression, []error) {
...
+
+		if len(err) > 0 {
+			return nil, append(errs, err...)
+		}
+	}
+	return expression, nil
 }
 
-// TODO: implement me w/ test cases :-)
 func parseInfixExpression(buf TokenBuffer, exp ast.Expression) (ast.Expression, []error) {
```
zeroFruit: I think parameter names with `left` instead of `exp` can be more intuitive

parse/parser.go
zeroFruit: Indentation quite unreadable... how about
```
expression := &ast.InfixExpression{
    Left: ...
}

**Review by zeroFruit — CHANGES_REQUESTED** [review]

File: parse/parser.go [hunk 3]
```diff
@@ -219,12 +249,11 @@ func parseIdentifier(buf TokenBuffer) (ast.Expression, []error) {
 		errs = append(errs, errors.New("parseIdentifier() - "+token.Val+" is not a identifier"))
 		return nil, errs
 	}
-	buf.Read()
 	return &ast.Identifier{Value: token.Val}, nil
 }
 
 func parseIntegerLiteral(buf TokenBuffer) (ast.Expression, []error) {
-	token := buf.Read()
+	token := buf.Peek(CURRENT)
```
zeroFruit: I think `Read()` is write. because when we execute this parse function we assume that current token is `IntegerLiteral`

parse/parser.go
```diff
@@ -200,16 +202,44 @@ func makePrefixExpression(buf TokenBuffer) (ast.Expression, []error) {
...
-	return nil, nil
+	errs := make([]error, 0)
+	expression := exp
+	curTok := buf.Peek(CURRENT)
+
+	for pre < nextPrecedence(buf) {
+		fn := infixParseFnMap[buf.Peek(NEXT).Type]
+
+		if fn == nil {
+			errs = append(errs, parseError{curTok.Type, "infix parse function not defined"})
```
zeroFruit: We don't need `curTok` except for this error message. so how about remove declaration curTok in line 208? And just `buf.Peek(CURRENT)` here

parse/parser.go
```diff
@@ -200,16 +202,44 @@ func makePrefixExpression(buf TokenBuffer) (ast.Expression, []error) {
...
-// then parse expression with that function if exist.
 func makeInfixExpression(buf TokenBuffer, exp ast.Expression, pre precedence) (ast.Expression, []error) {
-	return nil, nil
+	errs := make([]error, 0)
+	expression := exp
+	curTok := buf.Peek(CURRENT)
+
+	for pre < nextPrecedence(buf) {
+		fn := infixParseFnMap[buf.Peek(NEXT).Type]
+
```
zeroFruit, line 228: This is code style issue but I think this is quite eyesore to me.. Coding error handling right after definition is more readable:
```
fn := infixParseFnMap[buf.Peek(NEXT).Type]

if fn == nil {
```
to
```
fn := infixParseFnMap[buf.Peek(NEXT).Type]
if fn == nil {
```

parse/parser.go
```diff
@@ -200,16 +202,44 @@ func makePrefixExpression(buf TokenBuffer) (ast.Expression, []error) {
 	return exp, errs
 }
 
-// TODO: implement me w/ test cases :-)
-// MakeInfixExpression retrieves infix parse function from map
-// then parse expression with that function if exist.
```
zeroFruit: And plz remain comments except TODO: comments

parse/parser.go
```diff
@@ -157,9 +162,6 @@ func parseStatement(buf TokenBuffer) (ast.Statement, []error) {
 	}
 }
 
-// TODO: create test cases :-)
-// ParseExpression parse expression in two ways, first
-// by considering expression as prefix, next as infix
```
zeroFruit: And plz remain comments except TODO: comments

parse/parser.go
```diff
@@ -104,11 +104,16 @@ func expectNext(buf TokenBuffer, t TokenType) (bool, error) {
 }
 
 func curPrecedence(buf TokenBuffer) precedence {
-	return precedenceMap[buf.Peek(CURRENT).Type]
+	if p, ok := precedenceMap[buf.Peek(CURRENT).Type]; ok {
```
zeroFruit, line 123: I think this is not needed. first checking existence everytime is quite troublesome... and second if value don't exist than map automatically return 0 and this can be working as LOWEST

ast/ast.go
```diff
@@ -127,7 +145,7 @@ type IntegerLiteral struct {
 func (il *IntegerLiteral) produce() {}
 
 func (il *IntegerLiteral) String() string {
-	return string(il.Value)
+	return strconv.FormatInt(il.Value, 10)
```
zeroFruit, line 135: 👍 you find the bug!

parse/parser.go
boohyunsik: > I think this is not needed. first checking existence everytime is quite troublesome... and second if value don't exist than map automatically return 0 and this can be working as LOWEST

LOWEST is 1.. So I added this code. But, i think there are other ways like adding Int, String... in precedence map.

ast/ast.go
boohyunsik: > 👍 you find the bug!

It was modified some days ago :-D

parse/parser.go
boohyunsik: Ok, Instead of, I should remove buf.Read() in makeInfixExpression()

boohyunsik:
I will make version which uses Read() in parseInteger(Boolean, String) only.

zeroFruit:
Ok solved the problem.

1. First, test case `TestMakeInfixExpression` is a little bit incorrect. let's assume that we parse '1 + 2 + 3' using `parseExpression` function. Then in `makePrefixExpression` we parse '1' using `parseIntegerLiteral`. In `parseIntegerLiteral` function we **`Read`** the '1' from the buffer then in the buffer there's ' + 2 + 3' left. **So the first token (more exactly saying, first parsed tokens using prefix function) should not be in the buffer**. So the tc should changed from
```
bufs := [][]Token{
		{{Type: Int, Val: "1"}, {Type: Plus, Val: "+"}, {Type: Int, Val: "2"}, {Type: Asterisk, Val: "*"},
```
to this
```
bufs := [][]Token{
		{{Type: Plus, Val: "+"}, {Type: Int, Val: "2"}, {Type: Asterisk, Val: "*"},
```

2. Found the way to `makeInfixExpression`, `parseInfixExpression` correctly, still we keep the rule to **reading part should be happened only in the parsing function, not in the make function**. 
The idea was **from 1** actually, what I mean is in `makeInfixExpression` **we already `Read()` whatever tokens except operator** so the pointer in the buffer always pointing to operator. as a result, we shouldn't read next token's precedence we should read current one.

and I *prototyped* like this.(these are passed test cases)
```go
func makeInfixExpression(buf TokenBuffer, exp ast.Expression, pre precedence) (ast.Expression, []error) {
	expression := exp

	for pre < curPrecedence(buf) {
		fn := infixParseFnMap[buf.Peek(CURRENT).Type]
		expression, _ = fn(buf, expression)
	}
	return expression, nil
}

func parseInfixExpression(buf TokenBuffer, left ast.Expression) (ast.Expression, []error) {
	tok := buf.Read()

	exp := &ast.InfixExpression{
		Operator: ast.Operator{
			Type: operatorTypeMap[tok.Type],
			Val: ast.OperatorVal(tok.Val),
		},
		Left: left,
	}

	precedence := precedenceMap[tok.Type]
	exp.Right, _ = parseExpression(buf, precedence)

	return exp, nil
}
```

ast/ast.go
```diff
@@ -76,6 +78,22 @@ type Operator struct {
...
+	"*":  Asterisk,
+	"/":  Slash,
+	"%":  Mod,
+	"<":  LT,
+	">":  GT,
+	"<=": LTE,
+	">=": GTE,
+	"==": EQ,
+	"!=": NOT_EQ,
+}
```
zeroFruit: How about move this map into `parse/parser.go` and instead of mapping `string` with `OperatorType`, `TokenType` with `OperatorType`? 

Advantage of this is that when we mapping `TokenType` and `OperatorType`, then this doesn't dependent on `string` value of token anymore

boohyunsik:
> Ok solved the problem.
> 
> 1. First, test case `TestMakeInfixExpression` is a little bit incorrect. let's assume that we parse '1 + 2 + 3' using `parseExpression` function. Then in `makePrefixExpression` we parse '1' using `parseIntegerLiteral`. In `parseIntegerLiteral` function we **`Read`** the '1' from the buffer then in the buffer there's ' + 2 + 3' left. **So the first token (more exactly saying, first parsed tokens using prefix function) should not be in the buffer**. So the tc should changed from
> 
> ```
> bufs := [][]Token{
> 		{{Type: Int, Val: "1"}, {Type: Plus, Val: "+"}, {Type: Int, Val: "2"}, {Type: Asterisk, Val: "*"},
> ```
> to this
> 
> ```
> bufs := [][]Token{
> 		{{Type: Plus, Val: "+"}, {Type: Int, Val: "2"}, {Type: Asterisk, Val: "*"},
> ```
> 1. Found the way to `makeInfixExpression`, `parseInfixExpression` correctly, still we keep the rule to **reading part should be happened only in the parsing function, not in the make function**.
>    The idea was **from 1** actually, what I mean is in `makeInfixExpression` **we already `Read()` whatever tokens except operator** so the pointer in the buffer always pointing to operator. as a result, we shouldn't read next token's precedence we should read current one.
> 
> and I _prototyped_ like this.(these are passed test cases)
> 
> ```go
> func makeInfixExpression(buf TokenBuffer, exp ast.Expression, pre precedence) (ast.Expression, []error) {
> 	expression := exp
> 
> 	for pre < curPrecedence(buf) {
> 		fn := infixParseFnMap[buf.Peek(CURRENT).Type]
> 		expression, _ = fn(buf, expression)
> 	}
> 	return expression, nil
> }
> 
> func parseInfixExpression(buf TokenBuffer, left ast.Expression) (ast.Expression, []error) {
> 	tok := buf.Read()
> 
> 	exp := &ast.InfixExpression{
> 		Operator: ast.Operator{
> 			Type: operatorTypeMap[tok.Type],
> 			Val: ast.OperatorVal(tok.Val),
> 		},
> 		Left: left,
> 	}
> 
> 	precedence := precedenceMap[tok.Type]
> 	exp.Right, _ = parseExpression(buf, precedence)
> 
> 	return exp, nil
> }
> ```

Right, I also think we should remove Read() somewhere.(like parseInfix...)
There are so many Read() calls, so we feel difficulty thinking about parsing logic.

ast/ast.go
boohyunsik: Ok~

parse/parser.go
zeroFruit: ok on second thought checking map value might be good

parse/parser.go
```diff
@@ -200,26 +220,53 @@ func makePrefixExpression(buf TokenBuffer) (ast.Expression, []error) {
...
-func parseInfixExpression(buf TokenBuffer, exp ast.Expression) (ast.Expression, []error) {
-	return nil, nil
+func parseInfixExpression(buf TokenBuffer, left ast.Expression) (ast.Expression, []error) {
+	expression := &ast.InfixExpression{Left: left,
+		Operator: ast.Operator{Type: OperatorTypeMap[buf.Peek(CURRENT).Type],
+			Val: ast.OperatorVal(buf.Peek(CURRENT).Val)}}
+	precedence := precedenceMap[buf.Peek(CURRENT).Type]
+
+	buf.Read()
+
```
zeroFruit, line 255: How about `Read` the token from buffer before we define `expression`? then we don't need multiple `Peek`

parse/parser.go
zeroFruit: this is not fixed

parse/parser.go
```diff
@@ -61,6 +61,22 @@ var precedenceMap = map[TokenType]precedence{
 	Rbrace: LOWEST,
 }
 
+// Operator map
+var OperatorTypeMap = map[TokenType]ast.OperatorType{
```
zeroFruit: set private might be better

ast/ast.go
```diff
@@ -156,3 +158,16 @@ func (pe *PrefixExpression) produce() {}
...
+type InfixExpression struct {
+	Left Expression
+	Operator
+	Right Expression
+}
+
+func (ie *InfixExpression) produce() {}
+
+func (ie *InfixExpression) String() string {
+	return "(" + ie.Left.String() + string(ie.Operator.Val) + ie.Right.String() + ")"
```
zeroFruit: for readability how about put whitespace between characters? 
using `fmt.Sprintf` would make code pretty

parse/parser.go
```diff
@@ -200,26 +220,53 @@ func makePrefixExpression(buf TokenBuffer) (ast.Expression, []error) {
...
+		}
+	}
+	return expression, nil
 }
 
-// TODO: implement me w/ test cases :-)
-func parseInfixExpression(buf TokenBuffer, exp ast.Expression) (ast.Expression, []error) {
-	return nil, nil
+func parseInfixExpression(buf TokenBuffer, left ast.Expression) (ast.Expression, []error) {
+	expression := &ast.InfixExpression{Left: left,
```
zeroFruit: plz from
```
expression := &ast.InfixExpression{Left: left,
```
to
```
expression := &ast.InfixExpression{
    Left: left,
```

zeroFruit (changes requested):

boohyunsik:
Tomorrow i will make error test cases :-)
It is almost done!

PR #95, opened 2019-01-07
implement encodingBool

### LINKED ISSUE #84: [encode] Implement encodeBool
**Subject**
encode

**Detail**
```go
// EncodeOperand boolean to hexadecimal bytes
// ex) bool true => 01
// ex) bool false => 00
// TODO: implement w/ test cases :-)
func encodeBool(operand bool) ([]byte, error) {
	return nil, nil
}
```

Implement this with test case!

**Reference**
#29

geekseong:
resolved: #84 

details:
implement encodingBool

1. implement encodingBool
2. write down test case

- [ ] Test case


encoding/encode.go
```diff
@@ -66,13 +66,29 @@ func encodeInt(operand int) ([]byte, error) {
 // ex) string "abc" => 0x616263
 // TODO: implement w/ test cases :-)
```
AgwaB: plz remove TODO comment !

encoding/encode.go
```diff
@@ -66,13 +66,29 @@ func encodeInt(operand int) ([]byte, error) {
...
+		return nil, err
+	}
+
+	return dst, nil
 }
 
 // Encode boolean to hexadecimal bytes
 // ex) bool true => 0x01
 // ex) bool false => 0x00
 // TODO: implement w/ test cases :-)
```
AgwaB: here too

AgwaB:
LGTM

encoding/encode.go
geekseong: i removed it

encoding/encode.go
geekseong: i removed it

zeroFruit:
How about reset and commit --amend last 2 commits?
> * encodingString bug fix
> * remove TODO comment

encoding/encode_internal_test.go
```diff
@@ -53,3 +53,68 @@ func Test_encodeInt(t *testing.T) {
...
+	for i, test := range tests {
+		op := test.operand
+		bytecode, err := encodeString(op)
+
+		if err != nil {
+			t.Fatalf("test[%d] - encodeInt() had error. err=%v", i, err)
+		}
+
+		if !bytes.Equal(bytecode, test.expected) {
+			t.Errorf("test[%d] - encodeInt() result wrong. expected=%x, got=%x", i, test.expected, bytecode)
```
boohyunsik: In line 85, you use t.Fatalf. So I think it is better using same method.

encoding/encode_internal_test.go
```diff
@@ -53,3 +53,68 @@ func Test_encodeInt(t *testing.T) {
...
+	for i, test := range tests {
+		op := test.operand
+		bytecode, err := encodeBool(op)
+
+		if err != nil {
+			t.Fatalf("test[%d] - encodeInt() had error. err=%v", i, err)
+		}
+
+		if !bytes.Equal(bytecode, test.expected) {
+			t.Errorf("test[%d] - encodeInt() result wrong. expected=%x, got=%x", i, test.expected, bytecode)
```
boohyunsik: Same as above comment.

junbeomlee:
Size of the encoded data should be 4bytes!!

AgwaB:
@junbeomlee Are you talking except a string, right?

junbeomlee:
Yes @leesd556

geekseong:
why dose it return 4 bytes?

boohyunsik:
Maybe guys don't like commit message like 'merge...'.
Plz change commit msg :-D

junbeomlee (approved):
Good work!

Issue #103, opened 2019-01-08
[vm] Implement opcode NOT


AgwaB:
**Detail**
Implement opcode NOT

**Reference**
https://github.com/ethereum/go-ethereum/blob/master/core/vm/instructions.go

Issue #102, opened 2019-01-08
[vm] Implement opcode EQ


AgwaB:
**Detail**
Implement opcode EQ(Equal)

**Reference**
https://github.com/ethereum/go-ethereum/blob/master/core/vm/instructions.go

Issue #101, opened 2019-01-08
[vm] Implement opcode GT


AgwaB:
**Detail**
Implement opcode GT(Greater Than)

**Reference**
https://github.com/ethereum/go-ethereum/blob/master/core/vm/instructions.go

PR #106, opened 2019-01-09
Implement opcode NOT/EQ/GT/LT w/ testcase

### LINKED ISSUE #100: [vm] Implement opcode LT
**Detail**
Implement opcode LT(Less Than)

**Reference**
https://github.com/ethereum/go-ethereum/blob/master/core/vm/instructions.go

joomanzi:
resolved: #100 #101 #102 #103 

details:
Implement opcode NOT/EQ/GT/LT w/ testcase

+ change item type. uint32 to int32
+ comment: in case of OpDiv and OpMod, we use euclidian div instead standard div
ex) -15 / 6 = -3 ( not -2)
- [ ] Test case

### DIFF
--- a/vm/vm.go
+++ b/vm/vm.go
@@ -17,9 +17,8 @@
 package vm
 
 import (
-	"errors"
-
 	"encoding/binary"
+	"errors"
 
 	"github.com/DE-labtory/koa/opcode"
 )
@@ -76,12 +75,7 @@ func (add) Do(stack *stack, _ asmReader) error {
 	y := stack.pop()
 	x := stack.pop()
 
-	X := int(x)
-	Y := int(y)
-
-	item := item(X + Y)
-
-	stack.push(item)
+	stack.push(x + y)
 
 	return nil
 }
@@ -94,12 +88,7 @@ func (mul) Do(stack *stack, _ asmReader) error {
 	y := stack.pop()
 	x := stack.pop()
 
-	X := int(x)
-	Y := int(y)
-
-	item := item(X * Y)
-
-	stack.push(item)
+	stack.push(x * y)
 
 	return nil
 }
@@ -112,12 +101,7 @@ func (sub) Do(stack *stack, _ asmReader) error {
 	y := stack.pop()
 	x := stack.pop()
 
-	X := int(x)
-	Y := int(y)
-
-	item := item(X - Y)
-
-	stack.push(item)
+	stack.push(x - y)
 
 	return nil
 }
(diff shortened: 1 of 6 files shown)


junbeomlee (approved):
Good work!

AgwaB:
Sorry, i know it `NOT` has something wrong.

@junbeomlee 
Is it right that `NOT` operator act as invert value?
ethereum uses `big.Not` and it doesn't act as `!` but act as `~`

joomanzi:
checked

vm/vm.go
```diff
@@ -157,35 +157,74 @@ func (mod) hex() []uint8 {
...
 	return []uint8{uint8(opcode.GT)}
 }
 
-// TODO: implement me w/ test cases :-)
 func (eq) Do(stack *stack, _ asmReader) error {
+	y := stack.pop()
+	x := stack.pop()
+
+	Y := int(y)
+	X := int(x)
```
boohyunsik: There are 3 code blocks exactly same as above 4 lines.
How about use function to reduce duplicate codes?

vm/vm.go
```diff
@@ -157,35 +157,74 @@ func (mod) hex() []uint8 {
...
 func (eq) hex() []uint8 {
 	return []uint8{uint8(opcode.EQ)}
 }
 
-// TODO: implement me w/ test cases :-)
 func (not) Do(stack *stack, _ asmReader) error {
+	x := stack.pop()
+
+	X := int(x)
+
```
boohyunsik, line 194: Or, Like this, (I think) we can make function which pops just one element is stack, and returns after converting int.

vm/vm.go
```diff
@@ -157,35 +157,74 @@ func (mod) hex() []uint8 {
...
+	// ~x = - x - 1
```
AgwaB: Expression of `~x` in go is `^x`. how about `^X`?

vm/vm.go
AgwaB: I think we can make function which convert `item` to `int` in stack.go
and also i think we should convert `item` to `int32` in above function. because `item` is uint32 and golang recognize the `int` or `uint` types as 32 bits in 32-bit system, 64 bits in 64-bit system. 
the reason why convert to `int32` is that if when we convert `uint32` to `int64` and `uint32` has negative number, `int64` can't recognize the negative number.

vm/vm.go
joomanzi: thanks. maybe that capital X and small x could be vague naming.
so how about change into this form. I referenced go-ethereum code.
```
x, y := int(stack.pop()), int(stack.pop())
```
@leesd556

vm/vm.go
joomanzi: confirm thx 👍

junbeomlee:
How about adding compare method!
```
func (i item) compare(item item){
    x, y: = int32(i), int32(item)
    if x < y {
        return xxx
    }else if x > y {
       ...
    }else{
       ...
    }
}
```

joomanzi:
well this is just my opinion, make compare function might bad.
1. Stack Package, where Item type is in, should only contain stack-related method
2. item holds not only int, but any other etc,. so this method could make error when do type-conversion or bad-experience

junbeomlee (approved):
Good!

PR #113, opened 2019-01-09
Implement IfStatement, BlockStatement

boohyunsik:
resolved: #[108]

WIP : Implement IfStatement, BlockStatement

- [x] Test case

### DIFF
--- a/ast/ast.go
+++ b/ast/ast.go
@@ -158,7 +158,11 @@ type IfStatement struct {
 func (is *IfStatement) do() {}
 
 func (is *IfStatement) String() string {
-	return fmt.Sprintf("if ( %s ) { %s }", is.Condition.String(), is.Consequence.String())
+	if is.Alternative == nil {
+		return fmt.Sprintf("if ( %s ) { %s }", is.Condition.String(), is.Consequence.String())
+	}
+	return fmt.Sprintf("if ( %s ) { %s } else { %s }", is.Condition.String(), is.Consequence.String(),
+		is.Alternative.String())
 }
 
 // Represent block statement
@@ -169,11 +173,11 @@ type BlockStatement struct {
 func (bs *BlockStatement) do() {}
 
 func (bs *BlockStatement) String() string {
-	var str string
+	str := make([]string, 0)
 	for i := range bs.Statements {
-		str = str + bs.Statements[i].String()
+		str = append(str, bs.Statements[i].String())
 	}
-	return str
+	return strings.Join(str, "/")
 }
 
 // Represent string literal
--- a/ast/ast.go
+++ b/ast/ast.go
@@ -148,6 +148,34 @@ func (rs *ReturnStatement) String() string {
 	return fmt.Sprintf("return %s", rs.ReturnValue.String())
 }
 
+// Represent if statement
+type IfStatement struct {
+	Condition   Expression
+	Consequence *BlockStatement
+	Alternative *BlockStatement
+}
+
+func (is *IfStatement) do() {}
+
+func (is *IfStatement) String() string {
+	return fmt.Sprintf("if ( %s ) { %s }", is.Condition.String(), is.Consequence.String())
+}
+
+// Represent block statement
+type BlockStatement struct {
+	Statements []Statement
+}
+
+func (bs *BlockStatement) do() {}
+
+func (bs *BlockStatement) String() string {
+	var str string
+	for i := range bs.Statements {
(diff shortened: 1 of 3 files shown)


parse/parser_internal_test.go
```diff
@@ -611,3 +611,45 @@ func TestParseInfixExpression(t *testing.T) {
...
+			{Type: Lparen, Val: "("},
+			{Type: Bool, Val: "true"},
+			{Type: Rparen, Val: ")"},
+			{Type: Lbrace, Val: "{"},
+			{Type: IntType, Val: "int"},
+			{Type: Ident, Val: "a"},
+			{Type: Rbrace, Val: "}"},
+		},
+	}
+
```
junbeomlee, line 1207: i think we should test `else` part also!

boohyunsik: Yeap. This part was paused. I will do it.

ast/ast.go
```diff
@@ -113,6 +113,34 @@ func (as *AssignStatement) String() string {
...
+type IfStatement struct {
+	Condition   Expression
+	Consequence *BlockStatement
+	Alternative *BlockStatement
+}
+
+func (is *IfStatement) do() {}
+
+func (is *IfStatement) String() string {
+	return fmt.Sprintf("if ( %s ) { %s }", is.Condition.String(), is.Consequence.String())
```
zeroFruit: we need to print else part if Alternative exist

parse/parser.go
```diff
@@ -337,3 +337,67 @@ func parseAssignStatement(buf TokenBuffer) (*ast.AssignStatement, []error) {
 func parseCallExpression(buf TokenBuffer, fn ast.Expression) (ast.Expression, []error) {
 	return nil, nil
 }
+
+func parseIfStatement(buf TokenBuffer) (*ast.IfStatement, []error) {
+	errs := make([]error, 0)
+	var err []error
+	tok := buf.Read()
+	if tok.Type != If {
+		errs = append(errs, errors.New("parseIfStatement() error - No IF statement"))
```
zeroFruit: How about using parseError instead of just errors.New?

boohyunsik: This will be fixed.

ast/ast.go
boohyunsik: Ok.

boohyunsik:
Plz review it! Parsing is almost done :-D

parse/parser.go
```diff
@@ -178,8 +178,18 @@ func Parse(buf TokenBuffer) (*ast.Program, error) {
 // TODO: implement me w/ test cases :-)
 func parseStatement(buf TokenBuffer) (ast.Statement, error) {
 	switch buf.Peek(CURRENT).Type {
+	case Assign:
```
zeroFruit: I think Assign should be dealt inside the parseAssignStatement, so I think this is not needed

parse/parser.go
```diff
@@ -178,8 +178,18 @@ func Parse(buf TokenBuffer) (*ast.Program, error) {
...
+	case BoolType:
+		return parseAssignStatement(buf)
+	case If:
+		return parseIfStatement(buf)
+	case IntType:
+		return parseAssignStatement(buf)
 	case Return:
 		return parseReturnStatement(buf)
+	case StringType:
+		return parseAssignStatement(buf)
```
zeroFruit: Why don't we sort cases IntType, BoolType, StringType first. I mean sort Type cases first

parse/parser.go
```diff
@@ -454,3 +464,83 @@ func isDataStructure(tok Token) bool {
 		tok.Type == IntType ||
 		tok.Type == BoolType
 }
+
+func parseIfStatement(buf TokenBuffer) (*ast.IfStatement, error) {
+	tok := buf.Read()
+	if tok.Type != If {
+		err := parseError{
+			tok.Type,
+			fmt.Sprintf("is not a If"),
```
zeroFruit: we can just use "" string, if we don't use variable in the string

parse/parser.go
boohyunsik: Oh, I sorted alphabetically.
I will modify it.

parse/parser.go
boohyunsik: But without this, there was an error.
 I will retry.

zeroFruit:
LGTM

zeroFruit (approved):

Issue #99, opened 2019-01-08
[parser] remove errors.New and refactor it to parseError


zeroFruit:
```
errors.New("parseIdentifier() - "+token.Val+" is not a identifier")
```
like this. code which use errors.New should be refactored using parseError

queuestack:
I'll work on this issue.

zeroFruit:
This issue is done in #118, closed

PR #118, opened 2019-01-11
Changed parsing function's return type from []error to error

### LINKED ISSUE #111: [parser] Refactor parsing function's return type to just error
I think there's no need to consider to more than one error is returned from parsing function.

So refactor return type from `[]error` to `error`

zeroFruit:
resolved: #111, #99

Changed parsing function's return type from []error to error

This is my mistake when designing parser structure first. Just returning one error is enough.

- [x] Test case

### DIFF
--- a/parse/parser.go
+++ b/parse/parser.go
@@ -17,7 +17,6 @@
 package parse
 
 import (
-	"errors"
 	"fmt"
 	"strconv"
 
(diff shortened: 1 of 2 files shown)


junbeomlee:
That's the PR that I've been waiting for.

junbeomlee (approved):
Good job!

PR #124, opened 2019-01-11
Implement memory with testcase

AgwaB:
details:
After some review, i'll implement testcase...
Because the memory usage logic is more than I expected, so I think I need to change gradually.

[task detail...]
Implement memory

### DIFF
--- a/vm/memory.go
+++ b/vm/memory.go
@@ -1,0 +1,150 @@
+// Copyright 2015 The go-ethereum Authors
+// This file is part of the go-ethereum library.
+//
+// The go-ethereum library is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Lesser General Public License as published by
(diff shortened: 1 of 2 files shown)


vm/memory.go
```diff
@@ -0,0 +1,133 @@
...
+import (
+	"fmt"
+	"math/big"
+
+	"github.com/ethereum/go-ethereum/common/math"
+	"github.com/pkg/errors"
+)
+
+var ErrInvalidMemory = errors.New("Invalid memory reference")
+
```
junbeomlee, line 45: `"github.com/pkg/errors"` should be `"errors"`

junbeomlee:
Put DE-labtory license below the ethereum licence!

junbeomlee (approved):

junbeomlee:
Nice work!

PR #128, opened 2019-01-14
Add contract proto

AgwaB:
details:
Add contract proto

`Contract` contains information about the function that the user called.
I think we should talk more with the compiler... @hihiboss  @junbeomlee

### DIFF
--- a/vm/contract.go
+++ b/vm/contract.go
@@ -1,0 +1,34 @@
+/*
+ * Copyright 2018 De-labtory
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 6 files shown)


vm/contract.go
```diff
@@ -0,0 +1,29 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package vm
+
+type Contract struct {
```
hihiboss, line 19: If there are comments for `Code` and `Input`, it would be better.

AgwaB: I added that!

vm/vm_test.go
```diff
@@ -17,9 +17,10 @@
 package vm
 
 import (
-	"reflect"
 	"testing"
 
+	"reflect"
+
 	"github.com/DE-labtory/koa/opcode"
 	"github.com/stretchr/testify/assert"
```
joomanzi: still we use this lib?

AgwaB: Maybe i made a mistake... :( 
thx!!

PR #134, opened 2019-01-14
Fix condition of parseBooleanLiteral & Remove stretchr test module

### LINKED ISSUE #131: [parser] fix parseBooleanLiteral func
```
func parseBooleanLiteral(buf TokenBuffer) (ast.Expression, error) {
	token := buf.Read()

	if token.Type != Bool {
		return nil, parseError{
			token.Type,
			fmt.Sprintf("parseBooleanLiteral() error - %s is not bool", token.Val),
		}
	}
        ...
}
```
I think `token.Type != Bool` condition should be changed. In this condition we should check whether `token.Type` is `True` or `False` not `Bool`. In addition we don't need `Bool` type in TokenType

zeroFruit:
resolved: #131 

* Fix condition of `parseBooleanLiteral`  
* Removed `Bool` token type
* Remove test modules from `vm_test.go`

- [x] Test case

### DIFF
--- a/parse/parser.go
+++ b/parse/parser.go
@@ -319,7 +319,7 @@ func parseIntegerLiteral(buf TokenBuffer) (ast.Expression, error) {
 func parseBooleanLiteral(buf TokenBuffer) (ast.Expression, error) {
 	token := buf.Read()
 
-	if token.Type != Bool {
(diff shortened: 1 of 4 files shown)


boohyunsik (approved):
LGTM

PR #139, opened 2019-01-14
Fix expectNext when successfully done, buffer Read() called

boohyunsik:
I fixed expectNext to buffer read.
But I think that we should modify "Boolean token type" to "Bool".
(@zeroFruit fixed that "True" and "False")

Because,
1. I feel difficulty when fixing parseBooleanLiteral(). It check 2 boolean type.
2. Every variable like Int and String uses "Int", "String" to their token type.
I think "true", and "false" use just "Bool" type is make sense.
How about that?


zeroFruit (approved):
lgtm!

junbeomlee (approved):

PR #140, opened 2019-01-15
AST String() func, tc and changed AST node structure

### LINKED ISSUE #132: [ast] we need test cases of AST String()
actually lots of bugs are hide in `ast.go` String(). So I think we need test cases for `String()`

zeroFruit:
resolved: #132 

* Add AST String() tc
* Changed Operator struct
* Changed Datastructure struct
* Changed BooleanLiteral struct

- [x] Test case

### DIFF
--- a/ast/ast.go
+++ b/ast/ast.go
@@ -75,6 +75,21 @@ const (
 	NOT_EQ                // !=
 )
 
+var OperatorTypeMap = map[OperatorType]string{
+	Plus:     "PLUS",
+	Minus:    "MINUS",
+	Bang:     "BANG",
+	Asterisk: "ASTERISK",
+	Slash:    "SLASH",
+	Mod:      "MOD",
+	LT:       "LT",
+	GT:       "GT",
+	LTE:      "LTE",
+	GTE:      "GTE",
+	EQ:       "EQ",
+	NOT_EQ:   "NOT_EQ",
+}
+
 type OperatorVal string
 
 // Operator represent operator between expression
@@ -90,7 +105,7 @@ func (o Operator) String() string {
 type DataStructureType int
 
 func (dst DataStructureType) String() string {
-	return strconv.Itoa(int(dst))
+	return DataStructureTypeMap[dst]
 }
 
 const (
@@ -99,6 +114,12 @@ const (
 	String
 	Bool
 )
+
+var DataStructureTypeMap = map[DataStructureType]string{
+	Int:    "INT",
+	String: "STRING",
+	Bool:   "BOOL",
+}
 
 type DataStructureVal string
 
--- a/ast/ast.go
+++ b/ast/ast.go
@@ -57,25 +57,26 @@ func (i *Identifier) String() string {
 
 func (i *Identifier) produce() {}
 
-type OperatorType int
+// Operator represent operator between expression
+type Operator int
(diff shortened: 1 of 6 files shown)


ast/ast.go
```diff
@@ -57,64 +57,63 @@ func (i *Identifier) String() string {
...
+// DataStructure represent identifier's data structure
+// e.g. string, int, bool
+type DataStructure int
 
 const (
-	_ DataStructureType = iota
-	Int
-	String
-	Bool
+	_ DataStructure = iota
```
hihiboss, line 103: This looks better than `DataStructureType`!

ast/ast.go
```diff
@@ -202,18 +200,27 @@ func (il *IntegerLiteral) String() string {
...
 }
 
+type BooleanType int
+
+const (
+	False BooleanType = iota
+	True
+)
+
+var BooleanTypeMap = map[BooleanType]string{
```
hihiboss: Why did you create `BooleanTypeMap` without using string `true` or `false`?

parse/parser_internal_test.go
```diff
@@ -386,29 +386,28 @@ func TestParseBooleanLiteral(t *testing.T) {
...
-				t.Fatalf("test[%d] - TestParseBooleanLiteral() wrong result. expected=%s, got=%s",
-					i, test.expected, exp.String())
-			}
+		lit, ok := exp.(*ast.BooleanLiteral)
+		if err == nil && !ok {
+			t.Fatalf("test[%d] - TestParseBooleanLiteral() returned expression is not *ast.BooleanLiteral", i)
+		}
+
+		if err == nil && lit.String() != test.expected.String() {
+			t.Fatalf("test[%d] - TestParseBooleanLiteral() wrong result. expected=%s, got=%s",
```
hihiboss: How about `t.Fatalf("test[%d] - TestParseBooleanLiteral() wrong result. expected=\"%s\", got=\"%s\"", i, test.expected, lit.String())`?

boohyunsik: I also think it has more readability. (in console)

zeroFruit: 👍

ast/ast.go
zeroFruit: At first thought this map is for keeping consistency with data structure or operator. And booleanLiteral `String()` method become much simple.

But on second thought, this can be style issue and I think there's no answer :) what do you think? which one looks better?

boohyunsik: It is just my opinion, 'int' and 'string', 'boolean' is same context.
We don't create IntType, StringType and we don't implement those struct in IntegerLiteral, StringLiteral.
These, only have their value. So, why does booleanLiteral have its type value?
I think it would be more complicated.

hihiboss (approved):
LGTM!

ast/ast.go
```diff
@@ -210,10 +208,7 @@ type BooleanLiteral struct {
 func (bl *BooleanLiteral) produce() {}
 
 func (bl *BooleanLiteral) String() string {
-	if bl.Value {
-		return "true"
-	}
-	return "false"
+	return strconv.FormatBool(bl.Value)
```
boohyunsik, line 211: Good!

PR #141, opened 2019-01-16
Add pop operation

### LINKED ISSUE #125: [vm] Implement opcode Pop
Detail
Implement opcode Pop

Reference
https://github.com/ethereum/go-ethereum/blob/master/core/vm/instructions.go

hea9549:
resolved: #125 

details: implement pop opcode

I made a pop of vm. But I don't know if this is necessary.

- [x] Test case

### DIFF
--- a/vm/vm.go
+++ b/vm/vm.go
@@ -200,8 +200,8 @@ func (not) hex() []uint8 {
 	return []uint8{uint8(opcode.NOT)}
 }
 
-// TODO: implement me w/ test cases :-)
 func (pop) Do(stack *stack, _ asmReader) error {
+	_ = stack.pop()
 	return nil
 }
 
--- a/vm/asm.go
+++ b/vm/asm.go
@@ -26,6 +26,7 @@ var opCodes = map[opcode.Type]opCode{
 	opcode.Sub:  sub{},
 	opcode.Div:  div{},
 	opcode.Mod:  mod{},
+	opcode.Pop:  pop{},
 	opcode.Push: push{},
 	opcode.LT:   lt{},
 	opcode.GT:   gt{},
--- a/vm/vm_test.go
+++ b/vm/vm_test.go
@@ -350,9 +350,28 @@ func TestNOT(t *testing.T) {
 	}
 }
 
-// TODO: implement test cases :-)
 func TestPop(t *testing.T) {
+	testByteCode := makeTestByteCode(
+		uint8(opcode.Push), int32ToBytes(1),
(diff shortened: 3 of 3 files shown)


vm/vm.go
```diff
@@ -200,8 +201,11 @@ func (not) hex() []uint8 {
 	return []uint8{uint8(opcode.NOT)}
 }
 
-// TODO: implement me w/ test cases :-)
 func (pop) Do(stack *stack, _ asmReader) error {
+	if stack.len() == 0 {
+		return ErrStackEmpty
+	}
+	_ = stack.pop()
```
boohyunsik, line 204: if pop retuns err, why don't you return err?

hea9549: Do you mean stack.pop() throws an error? stack.pop() does not return an error.

boohyunsik: ah, I misunderstood :-D

boohyunsik (approved):
LGTM :-)

hihiboss:
Is it really fine just removing item??

@hea9549 @boohyunsik  @zeroFruit @leesd556 @junbeomlee

leesd556:
@hihiboss 
I think it should be discussed especially with the compiler!

1. Is it ok just pop without error handling?
 -> EVM don't handle that error. why?

2. Is it ok just removing item ?
 -> EVM uses `intPool`, a temporary stack storage. why?

I'll register it to issue!

AgwaB (changes requested):
@hihiboss 
I think it should be discussed especially with the compiler!

1. Is it ok just pop without error handling?
 -> EVM don't handle that error. why?

2. Is it ok just removing item ?
 -> EVM uses `intPool`, a temporary stack storage. why?

I'll register it to issue!

hihiboss:
@leesd556 

Ok. Thank you!

hea9549:
> @hihiboss
> I think it should be discussed especially with the compiler!
> 
> 1. Is it ok just pop without error handling?
>    -> EVM don't handle that error. why?
> 2. Is it ok just removing item ?
>    -> EVM uses `intPool`, a temporary stack storage. why?
> 
> I'll register it to issue!

1. I submit issue about this #150 

2. I found lot's of article about intpool. In my opinion, an intpool is simply used for reuse. also written in evm's intpool.go
```
// intPool is a pool of big integers that
// can be reused for all big.Int operations.
```
It is used to reuse objects without creating new ones.

AgwaB (approved):

PR #143, opened 2019-01-17
Fix error messages showing line and column

### LINKED ISSUE #133: [parser] Fix returning error format
In parser.go, formats of returning error are different each other.
I mean, some returning error statement is like below,
return fmt.Sprintf("....")
it is not containing "%s" or "%v".

boohyunsik:
resolved: #133 

Fix error messages showing their line and column.

### DIFF
--- a/parse/parser.go
+++ b/parse/parser.go
@@ -124,12 +124,12 @@ func nextTokenIs(buf TokenBuffer, t TokenType) bool {
 // type of token we want, and if true then return it
 // otherwise return with error
 func expectNext(buf TokenBuffer, t TokenType) error {
-	nextType := buf.Peek(CURRENT).Type
-	if nextType != t {
+	tok := buf.Peek(CURRENT)
+	if tok.Type != t {
 		return parseError{
-			nextType,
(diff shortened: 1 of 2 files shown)


zeroFruit:
LGTM

zeroFruit (dismissed):

zeroFruit (commented):
How about create new error type like `expectError` which contains `line`, `column` field.
```
type expectError struct {
    line int
    column int
}
```

junbeomlee:
How about add new attribute like `line`, `column` field.

```
type parseError struct {
	token Token
	expectedToken Token
	line int
	column int
}

func (e parseError) Error() string {
	return fmt.Sprintf("[line %d, column %d] expected [%s], but got [%s]",
				TokenTypeMap[e.token], TokenTypeMap[e.expectedToken]),					e.Line, e.Column)
}
``` 

Then we do not have to use fmt.Sprintf("[line %d, column %d] expected [%s], but got [%s]",
TokenTypeMap[e.token], TokenTypeMap[e.expectedToken]), every time!!
just return parseError is enough

boohyunsik:
I will do it!

zeroFruit (approved):

PR #159, opened 2019-01-20
Added String() in constant.go.

hihiboss:
resolved:

details:

Implement String() in opcode.

~- [ ] Test case~

### DIFF
--- a/opcode/constant.go
+++ b/opcode/constant.go
@@ -16,6 +16,8 @@
 
 package opcode
 
+import "errors"
+
 type Type uint8
 
 const (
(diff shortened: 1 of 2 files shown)


opcode/constant_test.go
```diff
@@ -0,0 +1,24 @@
...
+ * limitations under the License.
+ */
+
+package opcode_test
+
+import "testing"
+
+// TODO: implement test cases :-)
+func TestType_ToString(t *testing.T) {
+
```
boohyunsik, line 23: Are you implementing?

opcode/constant.go
```diff
@@ -152,3 +154,38 @@ const (
 	// memory[offset:offset+32] = value
 	Mstore Type = 0x23
 )
+
+// Change the bytecode of an opcode to string.
+func (p Type) ToString() (string, error) {
```
AgwaB: in golang, I know as `toSting()` should be `String()` !

opcode/constant_test.go
hihiboss: Nop! I'll leave it good first issue!

boohyunsik (approved):
Look great to me.
I will do compiler with you!

PR #169, opened 2019-01-20
Add abi skeleton

AgwaB:
details:
- abi
- method
- argument
- type

we need to define argument type which will used in koa

### DIFF
--- a/abi/abi.go
+++ b/abi/abi.go
@@ -16,7 +16,43 @@
 
 package abi
 
-import "github.com/DE-labtory/koa/crpyto"
+import (
+	"encoding/json"
+	"strings"
+
+	"github.com/DE-labtory/koa/crpyto"
+)
+
+type ABI struct {
+	Methods []Method
+}
+
+func New(abiJSON string) (ABI, error) {
+	reader := strings.NewReader(abiJSON)
+	dec := json.NewDecoder(reader)
+
+	var abi ABI
+	if err := dec.Decode(&abi); err != nil {
+		return ABI{}, err
+	}
+
+	return abi, nil
+}
+
+// UnmarshalJSON is implementation of json.Decoder's UnmarshalJSON
+func (abi *ABI) UnmarshalJSON(data []byte) error {
+	var methods []Method
+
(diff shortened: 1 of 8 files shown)


abi/abi.go
```diff
@@ -16,22 +16,31 @@
...
-// The implementation below is implemented for the abi spec of ethereum.
-// https://solidity.readthedocs.io/en/develop/abi-spec.html
+type ABI struct {
+	Methods []Method
+}
 
-// Encode abi parameters
-func Encode(params ...[]byte) []byte {
-	return nil
+func New(abiJSON string) (ABI, error) {
```
boohyunsik, line 30: Why don't you return *ABI?

abi/abi.go
```diff
@@ -16,22 +16,31 @@
...
+	reader := strings.NewReader(abiJSON)
+	dec := json.NewDecoder(reader)
+
+	var abi ABI
+	if err := dec.Decode(&abi); err != nil {
+		return ABI{}, err
```
boohyunsik, line 36: Can you return nil instead of ABI? (It is about return type)

abi/abi.go
AgwaB: ABI is just value and there is no change in ABI!

abi/abi.go
AgwaB: ok! there is no reason to return `ABI{}` instead of `nil` ... :)

abi/abi.go
```diff
@@ -16,22 +16,31 @@
...
+// The implementation below is implemented for the abi spec of ethereum.
+// https://solidity.readthedocs.io/en/develop/abi-spec.html
+
 // Encode abi parameters with function Selector
 func EncodeWithSelector(selector []byte, params ...[]byte) []byte {
 	return nil
 }
-
-// Get function selector(4bytes) from string of function signature
-func Selector(functionSignature string) []byte {
```
zeroFruit: Why did you removed this function? just my curiosity

junbeomlee (approved):

PR #172, opened 2019-01-20
Implement parseFunction, parameter and test cases

boohyunsik:
Implement parse function and parameter

- [x] Test case

### DIFF
--- a/ast/ast.go
+++ b/ast/ast.go
@@ -242,6 +242,35 @@ func (bl *BooleanLiteral) String() string {
 	return strconv.FormatBool(bl.Value)
 }
 
+// Represent Function expression
+type FunctionLiteral struct {
+	Parameters []*ParameterLiteral
+	Body       *BlockStatement
+}
+
+func (fl *FunctionLiteral) produce() {}
+
+func (fl *FunctionLiteral) String() string {
+	strs := make([]string, 0)
+	for _, parameter := range fl.Parameters {
+		strs = append(strs, parameter.String())
+	}
+	strs = append(strs, fl.Body.String())
+	return strings.Join(strs, ", ")
+}
+
+// Represent Function Parameter expression
+type ParameterLiteral struct {
+	Identifier *Identifier
+	Type       DataStructure
+}
+
+func (pl *ParameterLiteral) produce() {}
+
+func (pl *ParameterLiteral) String() string {
+	return fmt.Sprintf("Parameter : (Identifier: %s, Type: %s)", pl.Identifier.String(), pl.Type.String())
+}
+
 // Represent prefix expression
 type PrefixExpression struct {
 	Operator
--- a/ast/ast.go
+++ b/ast/ast.go
@@ -104,12 +104,14 @@ const (
 	IntType
 	StringType
 	BoolType
+	VoidType
 )
 
 var DataStructureMap = map[DataStructure]string{
 	IntType:    "int",
 	StringType: "string",
 	BoolType:   "bool",
+	VoidType:   "void",
 }
 
 func (ds DataStructure) String() string {
@@ -167,7 +169,7 @@ func (is *IfStatement) String() string {
 // e.g. func foo(int a) { ... }
 type FunctionLiteral struct {
 	Name       *Identifier
-	Parameters []*Identifier
+	Parameters []*ParameterLiteral
 	Body       *BlockStatement
 	ReturnType DataStructure
 }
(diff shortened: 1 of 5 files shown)


parse/parser.go
```diff
@@ -381,6 +381,77 @@ func parseStringLiteral(buf TokenBuffer) (ast.Expression, error) {
...
+			"token is not identifier",
+		}
+	}
+	lit.Name = &ast.Identifier{Value: tok.Val}
+
+	if err = expectNext(buf, Lparen); err != nil {
+		return nil, err
+	}
+
+	if lit.Parameters, err = parseFunctionParameters(buf); err != nil {
```
hea9549, line 404: I think it has some bug when function has no parameter, ex) func a () {}
After checking Lparen, parseFunctionParameters does not read Lparen from buf and checks Rparen, so it can not filter out empty parameters.

I ran the following test case (1) in your branch function TestParseFunctionLiteral and found the error (2).

(1)
``` go
{
	// fun example () {}
	{Type: Function, Val: "func"},
	{Type: Ident, Val: "example"},
	{Type: Lparen, Val: "("},
	{Type: Rparen, Val: ")"},
	{Type: Lbrace, Val: "{"},
	{Type: Rbrace, Val: "}"},
}
```

(2)
```
--- FAIL: TestParseFunctionLiteral (0.00s)
    parser_internal_test.go:522: test[1] - TestParseFunctionLiteral wrong result
        expected: func name(Parameter : (Identifier: a, Type: int), Parameter : (Identifier: b, Type: string))  {
        int [IDENT, c] = 5
        }
        got: func example()  {
        
        }
FAIL
```

hea9549 (changes requested):

parse/parser.go
hea9549: Therefore, I hope to add a TC for the empty parameter. :)

boohyunsik: Ok I will check it :-)

hea9549: no i think my mistake

hea9549 (approved):

parse/parser.go
hea9549: expectNext has buf.read().I didnt catch it.

parse/parser_internal_test.go
```diff
@@ -450,6 +452,131 @@ func TestParseStringLiteral(t *testing.T) {
 	}
 }
 
+func TestParseFunctionLiteral(t *testing.T) {
+	bufs := [][]Token{
```
zeroFruit, line 456: How about integrate bufs and tests in one struct? something like
```
struct {
   tokens
   expectedResult
}
```
tokens are part of test

zeroFruit: I think there's missing test cases which test
* empty parameters
* when there's return type

ast/ast.go
```diff
@@ -242,6 +211,47 @@ func (bl *BooleanLiteral) String() string {
...
+	out.WriteString(") ")
+	out.WriteString(fl.ReturnType.String() + " {\n")
+	out.WriteString(fl.Body.String() + "\n")
+	out.WriteString("}")
+
+	return out.String()
+}
+
+// Represent Function Parameter expression
+type ParameterLiteral struct {
```
zeroFruit, line 248: 👍

parse/parser.go
zeroFruit: And plz add comment of this function

parse/parser.go
```diff
@@ -381,6 +381,77 @@ func parseStringLiteral(buf TokenBuffer) (ast.Expression, error) {
...
+		return nil, err
+	}
+
+	if lit.Body, err = parseBlockStatement(buf); err != nil {
+		return nil, err
+	}
+	return lit, nil
+}
+
+func parseFunctionParameters(buf TokenBuffer) ([]*ast.ParameterLiteral, error) {
```
zeroFruit, line 427: plz add comment of this function

hea9549: when parameter has only ident , ex) func example(a a)
this function didn't return error.
I think it's a good idea to check error in parseFunctionParameters.

parse/parser.go
```diff
@@ -534,7 +605,7 @@ func parseIfStatement(buf TokenBuffer) (*ast.IfStatement, error) {
 	}
 
 	tok = buf.Read()
-	if curTokenIs(buf, Else) {
+	if err := expectNext(buf, Else); err == nil {
```
zeroFruit: @boohyunsik 
else is optional in if-else statements so I think it is better not to use `expect` function!

parse/parser.go
```diff
@@ -558,13 +629,7 @@ func parseBlockStatement(buf TokenBuffer) (*ast.BlockStatement, error) {
 		if stmt != nil {
 			block.Statements = append(block.Statements, stmt)
 		}
-		buf.Read()
+		//buf.Read()
```
zeroFruit: Why commented?

boohyunsik: Mistake ;-3

parse/parser_internal_test.go
boohyunsik: Yeap :-D

parse/parser.go
boohyunsik: ok

parse/parser.go
boohyunsik: Ok

parse/parser_internal_test.go
boohyunsik: > How about integrate bufs and tests in one struct? something like
> 
> ```
> struct {
>    tokens
>    expectedResult
> }
> ```
> tokens are part of test

For modifiying every test case, I will add issue.

zeroFruit: nic!

zeroFruit (approved):

PR #182, opened 2019-01-21
Implement parseFunctionStatement and test cases

### LINKED ISSUE #173: [ast, parser] Add ExpressionStatement and parsing function
**Subject**
ast, parser

**Detail**
This issue is related with #171 
function call without assignment should be treated as statement. and for this we need "expression statement"

so we need to add expression statement to ast and its parsing function.

```
func parseStatement(buf TokenBuffer) (ast.Statement, error) {
	switch tt := buf.Peek(CURRENT).Type; tt {
	case IntType:
		return parseAssignStatement(buf)
	case BoolType:
		return parseAssignStatement(buf)
	case StringType:
		return parseAssignStatement(buf)
	case If:
		return parseIfStatement(buf)
	case Return:
	default:
		// add parsing function here
	}
}
```

boohyunsik:
resolved: #173

1. Define FunctionStatemenet structure
2. Implement parseFunctionStatement()

- [x] Test case

### DIFF
--- a/parse/parser.go
+++ b/parse/parser.go
@@ -217,10 +217,7 @@ func parseStatement(buf TokenBuffer) (ast.Statement, error) {
 	case Return:
 		return parseReturnStatement(buf)
 	default:
-		return nil, parseError{
-			tokenType: tt,
-			reason:    "invalid token for parsing statement",
-		}
+		return parseExpressionStatement(buf)
 	}
 }
 
@@ -555,7 +552,10 @@ func parseCallExpression(buf TokenBuffer, fn ast.Expression) (ast.Expression, er
 
 func parseCallArguments(buf TokenBuffer) ([]ast.Expression, error) {
 	args := []ast.Expression{}
-	buf.Read()
+	if err := expectNext(buf, Lparen); err != nil {
+		return nil, err
+	}
+
 	if curTokenIs(buf, Rparen) {
 		return args, nil
 	}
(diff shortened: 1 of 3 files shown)


zeroFruit:
We need `ExpressionStatement` in ast, not `FunctionStatement`!
```
type ExpressionStatement struct {
	Expression
}

// And this is not expression but statement!
func (as *ExpressionStatement) do() {}
func (as *ExpressionStatement) String() string {}
```

zeroFruit:
We're trying to solve issue when function call exists sole, not assigned to identifier like:
```
int a = foo() <- expression
foo() // <- statement
```

boohyunsik:
> We're trying to solve issue when function call exists sole, not assigned to identifier like:
> 
> ```
> int a = foo() <- expression
> foo() // <- statement
> ```

Hmm, I intended it.

zeroFruit:
Oh yes. `FunctionStatement` do same thing but if we make `ExpressionStatement` all the expression can be statement.

parse/parser.go
```diff
@@ -216,6 +216,8 @@ func parseStatement(buf TokenBuffer) (ast.Statement, error) {
 		return parseIfStatement(buf)
 	case Return:
 		return parseReturnStatement(buf)
+	case Ident:
+		return parseExpressionStatement(buf)
 	default:
 		return nil, parseError{
```
zeroFruit: In default case we should `parseExpressionStatement`!

parse/parser.go
```diff
@@ -555,7 +552,10 @@ func parseCallExpression(buf TokenBuffer, fn ast.Expression) (ast.Expression, er
 
 func parseCallArguments(buf TokenBuffer) ([]ast.Expression, error) {
 	args := []ast.Expression{}
-	buf.Read()
+	if err := expectNext(buf, Lparen); err != nil {
+		return nil, err
+	}
+	//buf.Read()
```
zeroFruit: plz rm comment

zeroFruit (dismissed):

parse/parser.go
```diff
@@ -665,3 +665,23 @@ func parseBlockStatement(buf TokenBuffer) (*ast.BlockStatement, error) {
...
+func parseExpressionStatement(buf TokenBuffer) (*ast.ExpressionStatement, error) {
+	stmt := &ast.ExpressionStatement{}
+	tok := buf.Read()
+	if tok.Type != Ident {
+		return nil, parseError{
+			tok.Type,
+			"token is not identifier",
+		}
+	}
+	fn := &ast.Identifier{Value: tok.Val}
```
zeroFruit: fn -> ident might better!

zeroFruit (approved):

PR #191, opened 2019-01-24
Refactor parsePrefixExpression, block some expressions #189

### LINKED ISSUE #189: [parser] refactor parsePrefixExpression, block some expressions
**Subject**
parser

**Detail**
There's test code `TestMakePrefixExpression` in `parser_internal_test.go`. And we should block two cases.
```go
{
			tokens: []Token{
				{Type: Minus, Val: "-"},
				{Type: True, Val: "true"},
			},
			expected: "(-true)",
		},
		{
			tokens: []Token{
				{Type: Bang, Val: "!"},
				{Type: String, Val: "hello"},
			},
			expected: `(!"hello")`,
		},
```

## Cases
We should block these cases
1. -true like case
2. !"hello" like case

**And I think we should fix `parsePrefixExpression` function in `parser.go`**

u5surf:
resolved: #189 

details: I added validation rules to block prefix ‘!‘ is not valid for StringLiteral.
and ‘-’ is also invalid for Boolean.

[task detail...]

1. I added validations to block invalid prefix for some Literals.
2. I fix the TestMakePrefix

- [x] Test case
I fix TestMakePrefix to pass invalid error meesage to indicate blocking following #189 cases.

### DIFF
--- a/parse/parser.go
+++ b/parse/parser.go
@@ -379,6 +379,8 @@ func parsePrefixExpression(buf TokenBuffer) (ast.Expression, error) {
 			return nil, parseError{
 				tok.Type,
 				fmt.Sprintf("parsePrefixExpression() - Invalid prefix of %s", right.String()),
+				tok.Line,
+				tok.Column,
 			}
 		}
 	case ast.Minus:
@@ -387,6 +389,8 @@ func parsePrefixExpression(buf TokenBuffer) (ast.Expression, error) {
 			return nil, parseError{
 				tok.Type,
 				fmt.Sprintf("parsePrefixExpression() - Invalid prefix of %s", right.String()),
+				tok.Line,
+				tok.Column,
 			}
 		}
 	}
--- a/parse/parser.go
+++ b/parse/parser.go
@@ -18,7 +18,6 @@ package parse
 
 import (
 	"fmt"
-	"reflect"
 	"strconv"
 
 	"github.com/DE-labtory/koa/ast"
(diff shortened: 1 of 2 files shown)


zeroFruit (commented):
### First of all, thank you for contributing to our project! @u5surf  👍 

## Feedback

The logic is ok. But I think we can refactor and improve more.

1. **How about validating operation and right expression first, then create whole expression?**
The old code create whole expression first then filling the operation and right expression. I think after validating op and right expression then creates whole expression is more make sense.
```go
func parsePrefixExpression(buf TokenBuffer) (ast.Expression, error) {
	tok := buf.Read()

	op := operatorMap[tok.Type]
	right, err := parseExpression(buf, PREFIX)

        // validating logic

	// create whole expression
}
```

2. **We can use `switch` instead of `if`. I think using `switch` can be more readable and scalable.**
```go
switch op {
case ast.Bang:
    ...
case ast.Minus:
    ...
}
```
and
```go
switch right.(type) {
case *ast.BooleanLiteral:
    ...
case *ast.StringLiteral:    
    ...
}
```

junbeomlee:
Error has occurred !!
```
parse/parser.go:381:16: too few values in parseError literal
parse/parser.go:389:16: too few values in parseError literal
```
plz fix the error
you can check this in here https://travis-ci.org/DE-labtory/koa/builds/483891010?utm_source=github_status&utm_medium=notification

junbeomlee:
There is still an error in the code. 
Could u check parse/parser.go: 381 lines and
389 line!

u5surf:
Oh, I got it, parseError has added some element imperceptibly!

parse/parser.go
```diff
@@ -367,16 +367,38 @@ func parseInfixExpression(buf TokenBuffer, left ast.Expression) (ast.Expression,
 func parsePrefixExpression(buf TokenBuffer) (ast.Expression, error) {
 	var err error
```
zeroFruit, line 368: I think this `err` is not needed

Issue #186, opened 2019-01-22
[discuss] eol issue


joomanzi:
**Subject**
related to #138 
syntax rule : insert semicolon logic

**Proposal Detail**
We tried to make non-semicolon language for cleaner code, we need to define syntax in a concrete view
basically what I supposed to do is based on go language grammar . but we don't need to follow all the  rule while we are focusing on make more 'simple' language as a smart-contract language.

 so here is my proposal:
1) Specification
slightly different from go lang spec

a) When the input is broken into tokens, a semicolon is automatically inserted into the token stream at the end of a non-blank line if the line's final token

    - an identifier or basic literal
    - one of the tokens ++ -- ) }

b) To allow complex statements to occupy a single line, a semicolon may be omitted before a closing ) or }.

we don't support switch statement therefore we don't have to check keywords.(no keyword located at the end of line) 

2) Replacement
a) now we use semicolon( ";" ) for term as end of statement or expression, which won't appear on source code. 
b) EOL token which is  used for statement or expression just represent '\n' now . so delete EOL token

any suggestion about my proposal or something that i missed would welcome

Issue #138, opened 2019-01-14
[parser] add Eol check logic in the end of the line


junbeomlee:
**Where**
parse.go

**Bug detail**
We should check the end of line! 

**Fix plan**
Add following line `expectNext(buf,Eol)`

joomanzi:
this issue might handle in lexer

In golang, compiler automatically add semicolon at end of statement.
and by semicolon, easily check whether  token stream is at end of statement or not
we can follow this rule

check this out:
https://golang.org/ref/spec#Semicolons
https://golang.org/doc/effective_go.html#semicolons
https://medium.com/golangspec/automatic-semicolon-insertion-in-go-1990338f2649

joomanzi:
referece go_sdk src/go/scanner insertSemi

zeroFruit:
Really helpful!

zeroFruit:
on it! @zeroFruit

PR #193, opened 2019-01-24
add new operator and fix eol issue

### LINKED ISSUE #144: [lexer] New operator
**Subject**
Add new operator

**Detail**
- [ ] +=
- [ ] -=
- [ ] *=
- [ ] /=
- [ ] %=
- [ ] &&
- [ ] ||
- [ ] ++
- [ ] --

joomanzi:
resolved: #144 #138 #186

1. add new operator and refactoring defaultStateFn
+=, -=, *=, /=, %=, &&, ||, ++, --
2. Add insertSemi val and fix eol issue

add insertSemi value in stateFn and solve EOL issue in lexer.
now lexer don't return EOL token.
semicolon added when final token at end of line(\n) is

- Number, String, Ident
- ++, --
- } )
it consume every comment block or line
- [ ] Test case

### DIFF
--- a/parse/lex.go
+++ b/parse/lex.go
@@ -254,59 +254,89 @@ func (s *state) acceptRun(valid string) {
 //	Rbrace // }
 //
 
-func defaultStateFn(s *state, e emitter) stateFn {
+func (s *state) isNextToken(next rune) bool {
+	if s.peek() == next {
+		s.next()
+		return true
+	} else {
+		return false
+	}
+}
 
+func defaultStateFn(s *state, e emitter) stateFn {
 	switch ch := s.next(); {
 	case ch == '!':
-		if s.peek() == '=' {
-			s.next()
+		if s.isNextToken('=') {
 			e.emit(s.cut(NOT_EQ))
 		} else {
 			e.emit(s.cut(Bang))
 		}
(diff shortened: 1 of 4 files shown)


parse/lex.go
```diff
@@ -254,66 +255,100 @@ func (s *state) acceptRun(valid string) {
 //	Rbrace // }
 //
 
+func (s *state) isNextToken(next rune) bool {
+	if s.peek() == next {
+		s.next()
+		return true
+	} else {
+		return false
+	}
```
boohyunsik, line 262: How about think this?
```go
if s.peek() == next {
    s.next()
    return true
}
return false

File: parse/token.go [hunk 2]
```diff
@@ -111,6 +123,17 @@ var TokenTypeMap = map[TokenType]string{
 	Slash:    "SLASH",
 	Mod:      "MOD",
 
+	Plus_assign:     "PLUS_ASSIGN",
```
boohyunsik: In ast.go, we are using camel type (like PlusAssign).
I think we should use same type.
How about this?

joomanzi: appreciate u Mr.Boo

junbeomlee (approved):
LGTM!

PR #194, opened 2019-01-24
Add type checker for assignment statement

boohyunsik:
Add type checker in parseAssignmentState()

For example, if we input below sentence,
int a = "abcd"
This should fail unconditionally, but out parser was not.

So I added logic that compare identifier and assigner's type.

### DIFF
--- a/parse/parser.go
+++ b/parse/parser.go
@@ -49,6 +49,16 @@ var datastructureMap = map[TokenType]ast.DataStructure{
 	VoidType:   ast.VoidType,
 }
 
+// identifierTypeMap maps TokenType and TokenType.
+// It is used checking identifier and assigned check
+// for example, int a = "1234" will be failed.
+var identifierTypeMap = map[TokenType]ast.DataStructure{
+	Int:    ast.IntType,
+	True:   ast.BoolType,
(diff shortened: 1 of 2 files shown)


parse/parser.go
```diff
@@ -49,6 +49,16 @@ var datastructureMap = map[TokenType]ast.DataStructure{
 	VoidType:   ast.VoidType,
 }
 
+// identifierTypeMap maps TokenType and TokenType.
+// It is used checking identifier and assigned check
+// for example, int a = "1234" will be failed.
+var identifierTypeMap = map[TokenType]ast.DataStructure{
+	Int:    ast.IntType,
```
zeroFruit, line 56: Function Call can be assigned to int, string, bool type ident

boohyunsik: I asked you how can we implement this with function's return type.(I remember...maybe..)
We should have information of all functions before parsing, but It is very difficult.
Maybe this can be implemented after parsing (compile time)..
Or lexing phase, we can do something for this work.

boohyunsik: In this version, function just checks primitive type like int, string, bool.
so function call is not checked.

PR #195, opened 2019-01-25
Feature/compile expression

hihiboss:
resolved: 

details:

1. Added some functions for compiling expression.
2. Implemented Emit() function in bytecode.
3. Implemented compileBooleanLiteral with test cases.

- [x] Test case


translate/compiler.go
```diff
@@ -16,30 +16,41 @@
...
 		RawByte: make([]byte, 0),
 		AsmCode: make([]string, 0),
+		Abi: abi.ABI{
+			Methods: make([]abi.Method, 0),
+		},
 	}
 
 	for _, f := range c.Functions {
 		if err := compileFunction(*f, bytecode); err != nil {
 			return *bytecode, err
```
boohyunsik, line 42: Why return bytecode also?

translate/compiler_internal_test.go
```diff
@@ -16,7 +16,24 @@
...
+	"bytes"
+	"reflect"
+	"testing"
+
+	"github.com/DE-labtory/koa/abi"
+	"github.com/DE-labtory/koa/ast"
+)
+
+// TODO: implement test cases :-)
+func Test_generateFuncJumper(t *testing.T) {
```
boohyunsik: Hmm we need to talk about test function's name.
In parser, we named like TestGenerateFuncJumper()

translate/compiler_internal_test.go
```diff
@@ -29,6 +46,125 @@ func Test_compileStatement(t *testing.T) {
...
+	for i, test := range tests {
+		bytecode := &Bytecode{
+			RawByte: make([]byte, 0),
+			AsmCode: make([]string, 0),
+			Abi:     abi.ABI{},
+		}
+		err := compileBooleanLiteral(test.expression, bytecode)
+
+		if err != nil {
+			t.Fatalf("test[%d] - compileBooleanLiteral() had error. err=%v", i, err)
```
boohyunsik, line 145: We need to test whether function returns valid error or not.

boohyunsik (changes requested):

translate/compiler.go
hihiboss: @boohyunsik 
Your question means if error occurs, why return `*bytecode, err`, not `Bytecode{}, err`??
To be easier to debug or test, it returns the compiled bytecode so far.

translate/compiler_internal_test.go
hihiboss: Oh, i'll modify it!

translate/compiler_internal_test.go
hihiboss: But, the error is came out from just `encoding.EncodeOperand()`  not `compileBooleanLiteral()`. So, I think that's enough... Do you think it need?

boohyunsik: Anyway, the way for higher coverage is good. I think.

translate/compiler.go
```diff
@@ -106,3 +139,93 @@ func compileIfStatement(s *ast.IfStatement, bytecode *Bytecode) error {
...
+
+	case *ast.Identifier:
+		return compileIdentifier(expr, bytecode)
+
+	case *ast.ParameterLiteral:
+		return compileParameterLiteral(expr, bytecode)
+
+	default:
+		return nil
+	}
```
junbeomlee, line 186: should return errors when default!
```
default:
   return errors.New("error occurred")
```

AgwaB:
LGTM but you implement `Emerge` not `Emit` right?

zeroFruit:
After feedback the reviews which commented so far, looks good to me

hihiboss:
@leesd556 
Yes, just changed the name!
Emit means that function changes parameter like `Emit(bytecode)`
But bytecode will have that function, Emerge would be better. Like `Bytecode.Emerge()`

translate/compiler_internal_test.go
junbeomlee: we can not mock encoding function so i think this test case is enough!

junbeomlee:
Did u decide to add tracer?

PR #196, opened 2019-01-25
Add callfunc

AgwaB:
details:

- `asmReader` is a contract itself!
- `CallFunc` is just struct and has no any method because opCode will call function in `asmReader`!


hihiboss (approved):
I'm expecting next PR!

PR #200, opened 2019-01-26
refactor: appply extract function ParseError()

junbeomlee:
resolved: 

details:

1. refactor parse.go `return parseError{} ` to `return ParseError()`


hihiboss (approved):
LGTM

PR #202, opened 2019-01-26
Add parseExpectError for handling error in expectNext()

### LINKED ISSUE #198: [parser] Remove reason field in parseError
**Subject**
parser

**Detail**
```go
// ParseError contains error which happened during
// parsing tokens
type parseError struct {
	tokenType TokenType
	reason    string
	line      int
	column    Pos
}
```
Current `parseError` looks like this. But `reason` field make making `parseError` complex and code unreadable. So removing the `reason` field in `parseError` might make code better. 
Then create concrete parse error instead of comprehensive just `parseError` like:
```go
type parseExpectError struct {
}
type parseXXXError struct {
}
...
```

boohyunsik:
resolved: #198

Add parseExpectError for handling error in expectNext,
and I will fix all the remaining functions.

- [x] Test case

### DIFF
--- a/parse/parser.go
+++ b/parse/parser.go
@@ -126,11 +126,11 @@ func nextTokenIs(buf TokenBuffer, t TokenType) bool {
 func expectNext(buf TokenBuffer, t TokenType) error {
 	tok := buf.Peek(CURRENT)
 	if tok.Type != t {
-		return parseError{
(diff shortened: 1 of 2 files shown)


zeroFruit:
Nice work!

zeroFruit (approved):

PR #208, opened 2019-01-27
add land, lor parse function

### LINKED ISSUE #206: [parser] parse Land, Lor
**Subject**
parser

**Detail**
`&&`, `||` tokens are defined but there's no parsing function for those.

hea9549:
resolved: #206 

details:

add land, lor parse function

- [x] Test case

### DIFF
--- a/parse/parser.go
+++ b/parse/parser.go
@@ -38,6 +38,8 @@ var operatorMap = map[TokenType]ast.Operator{
 	GTE:      ast.GTE,
 	EQ:       ast.EQ,
 	NOT_EQ:   ast.NOT_EQ,
+	Land:     ast.LAND,
+	Lor:      ast.LOR,
 }
 
 // datastructureMap maps TokenType with Datastructure. By doing this
@@ -59,6 +61,8 @@ type precedence int
 const (
 	_ precedence = iota
 	LOWEST
+	LOR         // ||
+	LAND        // &&
 	EQUALS      // ==
 	LESSGREATER // > or <
 	SUM         // +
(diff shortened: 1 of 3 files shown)


parse/parser.go
```diff
@@ -38,6 +38,8 @@ var operatorMap = map[TokenType]ast.Operator{
 	GTE:      ast.GTE,
 	EQ:       ast.EQ,
 	NOT_EQ:   ast.NOT_EQ,
+	Land:	  ast.LAND,
```
boohyunsik: Should be LAND?

boohyunsik (changes requested):

parse/parser.go
hea9549: I make precedence name LAND and LOR.
So If I make Land to LAND, it has same name with precedence.
Also, other token name has lowercase like Slash, Mod, IntType and so on.
I think it is not bad to make name 'Land'.

parse/parser.go
```diff
@@ -59,6 +61,8 @@ type precedence int
 const (
 	_ precedence = iota
 	LOWEST
+	LOR         // ||
+	LAND        // &&
```
zeroFruit, line 65: I think LOR and LAND should have same precedence. plz add test case which contain both `||` and `&&`

for example
```
true || true && true // should ((true || true) && true)
```
```
true && true || true // should ((true && true) || true)
```

hea9549: complete > <

zeroFruit: my mistake 

https://golang.org/src/go/token/token.go

zeroFruit (approved):

PR #209, opened 2019-01-27
Implement default object structure

boohyunsik:
This implementation is for custom type and type checking

- [x] Test case

### DIFF
--- a/symbol/scope.go
+++ b/symbol/scope.go
@@ -16,6 +16,11 @@
 
 package symbol
 
+import (
+	"bytes"
+	"fmt"
+)
+
 // NewScope() makes a new scope.
 func NewScope() *Scope {
 	s := make(map[string]Symbol)
@@ -52,7 +57,16 @@ func (s *Scope) Set(name string, val Symbol) Symbol {
 	return val
 }
 
-// TODO implement me w/ test cases for debugging.
 func (s *Scope) String() string {
-	return ""
+	var out bytes.Buffer
+	scope := s
+	div := ""
+	for scope != nil {
+		out.WriteString(div)
+		out.WriteString("[ Scope ] ")
+		out.WriteString(fmt.Sprintln(scope.store))
+		div += "  "
+		scope = scope.outer
+	}
+	return out.String()
 }
--- a/symbol/scope.go
+++ b/symbol/scope.go
@@ -1,0 +1,58 @@
+/*
+ * Copyright 2018 De-labtory
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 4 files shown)


junbeomlee (approved):
Nice work!

zeroFruit:
How about using `Symbol` keyword instead of `Object`?

ref: [symbol table wiki](https://ko.wikipedia.org/wiki/%EC%8B%AC%EB%B3%BC_%ED%85%8C%EC%9D%B4%EB%B8%94)

The `Symbol` just need **name**, **type**, **scope**, not concrete info.

for example `Function` don't need to contain `Body` information. we need function *name*, and the type which tell us that *name* is function.

junbeomlee:
i agree with zeroFruit!

symbol/symbol.go
```diff
@@ -0,0 +1,90 @@
...
+)
+
+type Symbol interface {
+	Type() SymbolType
+	String() string
+}
+
+// Represent Integer symbol
+type Integer struct {
+	Value int64
```
zeroFruit: we don't need to save value, just integer variable name.
for example
```
int a = 1
```
in this case we need to save **'a' and the fact that 'a' is integer**, so `Integer` struct would be like this
```
type Integer struct {
    id Identifier
}

// usage
Integer{id: "a"}
```

symbol/scope.go
```diff
@@ -0,0 +1,58 @@
...
+func NewEnclosedScope(outer *Scope) *Scope {
+	scope := NewScope()
+	scope.outer = outer
+	return scope
+}
+
+// Scope represent a variable's scope.
+type Scope struct {
+	store map[string]Symbol
+	outer *Scope
```
zeroFruit, line 41: maybe we don't need `outer` for now. because our contract consists of functions, no nested function and block statement have same scope, i.e.
```
func foo() {
    // scope1

    if (expression) {
        // same scope with scope1
    }
}
```

symbol/symbol.go
boohyunsik: But in case of using a in other line, how can we inspect value of a?
i.e) int a = 1
a = a + 1

We have symbol table, so we can know a is integer.
But, we don't know value of a.
And all literals defined in ast package, have only their value except their name.
so I think we must have value of symbol in symbol table.

boohyunsik: Next, I think we don't need name value, because we use name as key of map.
like, map["a"] -> integer a

zeroFruit: > We have symbol table, so we can know a is integer.
But, we don't know value of a.
And all literals defined in ast package, have only their value except their name.
so I think we must have value of symbol in symbol table.

We don't need to know what value of 'a' is. just need to know what is identifier, what is expression. 
In the case of `a = a + 1`, what `Scope` should do is checking whether lhs `a` is defined before. 
* If not, return error. 
* If defined before, then parsing function builds AST as before (and this is parser part, not `Scope` part)

zeroFruit: > Next, I think we don't need name value, because we use name as key of map.
like, map["a"] -> integer a

Yes `name` can be redundant, but we use it later after retrieving symbol for validating. for example:

```
obj := scope.Get[token.Val]
if token.Val != obj.name { // we can explicitly compare with name
    ...
}
```

I think it is a little bit style issue :-)

zeroFruit (approved):
LGTM!! 💯

PR #212, opened 2019-01-28
Add opcode and callfunc method

AgwaB:
details:

1. Add opcode `loadfunc`, `loadargs`, `returning`
2. Add `CallFunc` structure's method : `function`, `arguments`

`return` is a keword so i named it as `returning`...

So, if you have any good idea about naming, plz recommend!


junbeomlee:
Nice work!

PR #222, opened 2019-01-30
Define MemoryTable.

hihiboss:
resolved: #[issue number]

details:

[task detail...]

1. [task1]
2. [task2]

- [ ] Test case

==

**I'm still thinking and designing now. This is just a concept.**
**Please comment any idea you have!**

### DIFF
--- a/translate/tracer.go
+++ b/translate/tracer.go
@@ -1,0 +1,41 @@
+/*
+ * Copyright 2018 De-labtory
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package translate
+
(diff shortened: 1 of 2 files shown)


translate/tracer.go
```diff
@@ -0,0 +1,41 @@
...
+package translate
+
+// Identifier saves size and offset of the value which the variable has.
+type Identifier struct {
+	Offset uint
+	Size   uint
+}
+
+// MemoryTable is used to know the location of the memory
+type MemoryTable struct {
```
zeroFruit, line 26: ~~I think in `MemoryTable`, we would save both identifier (variable name) and its value.~~ I think `MemoryTableEntry` would be better than `Identifier`.

ref:
* page table: https://www.geeksforgeeks.org/operating-system-page-table-entries/
* page table entry: https://www.geeksforgeeks.org/operating-system-page-table-entries/

I know that page table is not exactly same as memory table, but I posted the link because we may can use that words for similar concepts

translate/tracer.go
```diff
@@ -0,0 +1,41 @@
...
+	IdentifierMap map[string]Identifier
```
AgwaB: what is key(id) of `IdentifierMap`?
I think id will be stored in bytecode instead of offset and size. am i right?

translate/tracer.go
hihiboss: Thanks!

translate/tracer.go
hihiboss: The key is a variable's name. ID is not stored in bytecode. Offset and size would be stored in bytecode. For storing offset and size, we should know what is offset or size of the value. MemoryTable is used to know them.

zeroFruit (approved):

PR #225, opened 2019-01-31
test: add bool type test to parser integral test

zeroFruit:
details:

1. test bool as function return type	
2. test bool as function parameter	
3. test bool as variable type

- [x] Test case

### DIFF
--- a/parse/parser_test.go
+++ b/parse/parser_test.go
@@ -42,12 +42,6 @@ type parserTestCase struct {
 	}
 }
 
-/*
-TODO:
- 	1. test bool as function return type
-	2. test bool as function parameter
-	3. test bool as variable type
-*/
 func TestParse(t *testing.T) {
 	input := `
 	contract {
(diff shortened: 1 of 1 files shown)


parse/parser_test.go
```diff
@@ -91,8 +85,8 @@ func TestParse(t *testing.T) {
 				"hello"
 			string tabbed_string = "hello, \t world"
 
-			//bool a = true
-			//bool b= false
+			bool a = true
+			bool b= false
```
boohyunsik, line 89: space!

boohyunsik (changes requested):

parse/parser_test.go
zeroFruit: I skipped space on purpose :) to check whether no space work correctly!

boohyunsik: oh I know.

boohyunsik (approved):

Issue #201, opened 2019-01-26
[vm] Suggest abi encoding format


AgwaB:
**Subject**
abi.go

**Proposal Detail**
Implementation on #199 , i want to suggest abi parameter encoding format.

- We have only one dynamic type : `string`
- In encoded bytes, bytes has a `pointer`, `size`, `value` like a below
![dd](https://user-images.githubusercontent.com/26357797/51785899-2920a500-21a0-11e9-93d2-0a9a66b0da0d.png)

Pointer : 4bytes
Size : 4bytes
Value : according to size 

Ethereum uses all 32bytes but i think 4bytes is enough for pointer and size.

Example)
function foo(int, string, int) 
parameter : (50, "HelloKOA", 256)

0x00 : 0000000C
0x04 : 00000011
0x08 : 0000001D
0x0C : 00000001
0x10 : 32
0x11 : 00000008
0x15 : 48656c6c6f4b4f41
0x1D: 00000002
0x21 : 0100

PR #226, opened 2019-01-31
Implements abi_encode

### LINKED ISSUE #199: [vm] Implement Encode function
**Subject**
abi.go

**Detail**
- Implement `Encode` function
- koa's dynamic data type is only `string`
- string is right-padded
- others is left-padded

* You shoud know it. We are a little different from etherium. Reference is just reference

1. first spot is pointer to value's place.
2. jump to the point you met in 1
3. than you'll meet value's size
4. next point is value.

example
function foo(int, int, string)
parameter : (50, 256, "Hello Koa")
  0x00 : 0000000000000000000000000000000000000000000000000000000000000060
  0x20 : 00000000000000000000000000000000000000000000000000000000000000A0
  0x40 : 00000000000000000000000000000000000000000000000000000000000000E0
  0x60 : 0000000000000000000000000000000000000000000000000000000000000001
  0x80 : 0000000000000000000000000000000000000000000000000000000000000032
  0xA0 : 0000000000000000000000000000000000000000000000000000000000000002
  0xC0 : 0000000000000000000000000000000000000000000000000000000000000100
  0xE0 : 0000000000000000000000000000000000000000000000000000000000000009
0x100 : 48656c6c6f204b6f610000000000000000000000000000000000000000000000

**Reference**
solidity abi spec : https://solidity.readthedocs.io/en/v0.4.24/abi-spec.html
abi encoding simulator : https://adibas03.github.io/online-ethereum-abi-encoder-decoder/#/

AgwaB:
resolved: #199, #201 

details:

1. Implments `Encode` function
2. I modified encoding/encode.go function to use this function to `Encode` function

- [x] Test case

### DIFF
--- a/abi/abi_encode.go
+++ b/abi/abi_encode.go
@@ -1,0 +1,115 @@
+/*
+ * Copyright 2018 De-labtory
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package abi
+
+import (
+	"github.com/DE-labtory/koa/crpyto"
+	"github.com/DE-labtory/koa/encoding"
+)
+
+type Pointer []byte
+type Size []byte
+type Value []byte
+
+// Encode abi parameters
+func Encode(params ...interface{}) ([]byte, error) {
+	values, err := encodeValues(params...)
+	if err != nil {
+		return nil, err
+	}
+
+	sizes, err := encodeSizes(values)
+	if err != nil {
+		return nil, err
+	}
+
+	pointers, err := encodePointers(sizes, values)
+	if err != nil {
+		return nil, err
+	}
+
+	var Args []byte
+
+	for _, pointer := range pointers {
+		Args = append(Args, pointer...)
+	}
+
+	for index := 0; index < len(params); index++ {
+		Args = append(Args, sizes[index]...)
+		Args = append(Args, values[index]...)
+	}
+
+	return Args, nil
+}
+
+func encodeValues(params ...interface{}) ([]Value, error) {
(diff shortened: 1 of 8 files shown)


abi/abi_encode.go
```diff
@@ -0,0 +1,186 @@
...
+	for _, pointer := range pointers {
+		Args = append(Args, pointer...)
+	}
+
+	for index := 0; index < len(params); index++ {
+		Args = append(Args, sizes[index]...)
+		Args = append(Args, values[index]...)
+	}
+
+	return Args, nil
```
boohyunsik, line 56: Encode is so big. Can you divide it?

hihiboss: It would be better to divide them by `for` statement.

Ex)
```go
func Encode(params ...interface{}) ([]byte, error) {
  pointers := make([]Pointer, len(params))
  sizes := make([]Size, len(params))
  values := make([]Value, len(params)) 

  encodeValue(sizes, values)

  encodePointer(pointers)

  var Args []byte

  for _, pointer := range pointers {
    Args = append(Args, pointer...)
  }

  for index := 0; index < len(params); index++ {
    Args = append(Args, sizes[index]...)
    Args = append(Args, values[index]...)
  }
  
  return Args, nil
}

**Comment by hihiboss** [comment]
@leesd556 
`abi/encode.go` looks better than `abi/abi_encode.go`

File: encoding/encode.go [hunk 2]
```diff
@@ -18,9 +18,19 @@ package encoding
...
 import (
 	"encoding/hex"
-	"errors"
 	"fmt"
 	"strconv"
+
+	"github.com/pkg/errors"
+)
+
+type convertFunc func(int64) string
```
zeroFruit: I think we can have better name than `convertFunc`

abi/abi_encode.go
```diff
@@ -0,0 +1,115 @@
...
+}
+
+func encodeValues(params ...interface{}) ([]Value, error) {
+	values := make([]Value, len(params))
+
+	for index, param := range params {
+		bytesValue, err := encoding.EncodeOperand(param, 0)
```
zeroFruit: we can use `NO_PADDING` const instead of just 0

junbeomlee (approved):
LGTM!

encoding/encode.go
junbeomlee: i agree!! it needs to be named more understandable!

abi/abi_encode.go
AgwaB: fixed it!

junbeomlee (approved):
Nice work!

PR #229, opened 2019-02-01
bug: consume semicolon

### LINKED ISSUE #228: [parser] Optionally check semicolon for each statement
**Where**
every statement parse function

**Bug detail**
now every statement parse function check semicolon at the end of statement

**Fix plan**
every statement parse function peek semicolon at the end of statement, if it exists accept it or ignore

zeroFruit:
resolved: #228 

details:

Everytime we've done parsing statement, consume semicolon which follows statements

- [x] Test case


parse/parser.go
```diff
@@ -514,14 +520,7 @@ func parseFunctionLiteral(buf TokenBuffer) (*ast.FunctionLiteral, error) {
 		return nil, err
 	}
 
-	// After exit from parseBlockStatement, consume right-brace
-	if err = expectNext(buf, Rbrace); err != nil {
```
boohyunsik, line 518: Left brace is parsed in function parsing phase, but Right brace is parsed in block parsing phase.
I think braces parsing should be done in one phase.

junbeomlee:
LGTM!

PR #230, opened 2019-02-01
Implements compileReturnStatement and test cases

### LINKED ISSUE #218: [compiler] Implement compileReturnStatement()
**Subject**
compiler

**Detail**
Implement `compileReturnStatement()` in `translate/compiler.go`

boohyunsik:
resolved: #218

Implements compileReturnStatement and test cases

- [x] Test case

### DIFF
--- a/translate/bytecode.go
+++ b/translate/bytecode.go
@@ -18,6 +18,7 @@ package translate
 
 import (
 	"fmt"
+	"strings"
 
 	"github.com/DE-labtory/koa/opcode"
 )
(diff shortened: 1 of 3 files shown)


hihiboss:
@boohyunsik 
Return value could be integer, string or void!
So, I think you should separate their type.

If return value is integer, you should return only that integer value!
If return value is string, you should return memory offset and size of that string.
If return value is void, you should return memory offset and size 0.
Otherwise (if return value is not integer, string or void, but expression), you should execute `compileExpression()` function.

boohyunsik:
> @boohyunsik
> Return value could be integer, string or void!
> So, I think you should separate their type.
> 
> If return value is integer, you should return only that integer value!
> If return value is string, you should return memory offset and size of that string.
> If return value is void, you should return memory offset and size 0.
> Otherwise (if return value is not integer, string or void, but expression), you should execute `compileExpression()` function.

But in compileExpression(), the cases of integer, string, boolean are already defined.

hihiboss:
When `compileStringExpression()` is merged to develop, add the test case for string. Then, this might be merged.

PR #238, opened 2019-02-02
refactor: remove ParseError(), remove line, col field from errors

### LINKED ISSUE #236: [parser] refactor parseError, parseExpectError
**Subject**
parser error structure

**Detail**
```
type parseError struct {
	tokenType TokenType
	reason    string
	line      int
	column    Pos
}
```

```
type parseExpectError struct {
	tokenType TokenType
	expected  TokenType
	column    Pos
	line      int
}
```
both error don't need to have column, line, just token is enough.

```
func ParseError(msg string, token Token) parseError {
	return parseError{
		token.Type,
		msg,
		token.Line,
		token.Column,
	}
}
```

And this function is unnecessary is we refactor error

```
func (e parseError) Error() string {
	return fmt.Sprintf("[line %d, column %d] %s", e.line, e.column, e.reason)
}
```
Finally parseError should print token which makes error

zeroFruit:
resolved: #236 

details:

* remove `ParserError`
* change parser error field (remove `col`, `line` fields)
* some change from parseError -> parseExpectError

- [x] Test case

### DIFF
--- a/parse/parser.go
+++ b/parse/parser.go
@@ -132,11 +132,9 @@ func nextTokenIs(buf TokenBuffer, t TokenType) bool {
 func expectNext(buf TokenBuffer, t TokenType) error {
 	tok := buf.Peek(CURRENT)
 	if tok.Type != t {
-		return parseExpectError{
-			tok.Type,
+		return ParseExpectError{
+			tok,
 			t,
-			tok.Column,
-			tok.Line,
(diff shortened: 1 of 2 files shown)


boohyunsik (approved):
^^b

hihiboss (approved):
LGTM :)

PR #240, opened 2019-02-02
feat: skel adding scope to parser

### LINKED ISSUE #239: [parser] Apply Scope to parse functions
**Subject**
parser

**Detail**
apply scope to parsing function to validate type check

**Reference**
#203, #165, #152

zeroFruit:
resolved: #239 

details:

1. feat: skel adding scope to parsing functions with its helper fns
2. feat: add updateScopeSymbol to parseFunctionLiteral w/ tcs
3. refactor: extract parseFunctionParameter function

- [x] Test case

### DIFF
--- a/symbol/scope.go
+++ b/symbol/scope.go
@@ -18,7 +18,6 @@ package symbol
 
 import (
 	"bytes"
-	"errors"
 	"fmt"
 	"sort"
 )
@@ -45,16 +44,16 @@ type Scope struct {
 // Getter return's variable's scope.
 // If a scope doesn't have a variable,
 // Program will search in outer scope.
-func (s *Scope) Get(name string) (Symbol, error) {
+func (s *Scope) Get(name string) Symbol {
 	var scope = s
 	for scope != nil {
 		obj, ok := scope.store[name]
 		if ok {
-			return obj, nil
+			return obj
 		}
 		scope = scope.outer
 	}
-	return nil, errors.New(fmt.Sprintf("%s is not defined", name))
+	return nil
 }
 
 // Setter set a variable to target scope's map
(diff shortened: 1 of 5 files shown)


symbol/scope.go
```diff
@@ -45,16 +44,16 @@ type Scope struct {
...
 	var scope = s
 	for scope != nil {
 		obj, ok := scope.store[name]
 		if ok {
-			return obj, nil
+			return obj
 		}
 		scope = scope.outer
 	}
-	return nil, errors.New(fmt.Sprintf("%s is not defined", name))
```
boohyunsik, line 57: Question) What is the advantages deleting error?

zeroFruit: I thought that when symbol does not exist in Map this is not an error, we need to add new symbol to the table.
It is an error when the symbol exist in the table with the same name.

boohyunsik: Hmm, why that case is not error..? If user use the symbol which is not defined in any scope, program must occur error. I think this error should be occurred in get function.
So we can show error msg like this symbol is not defined. how about it?
Can we make not defined error in other place?

zeroFruit: how about following go map convention: return symbol with bool value? like
```
func (s *Scope) Get(name string) (Symbol, bool) {
   ...
}
```
```
// caller
sym, ok := scope.Get("a")
if !ok {
   // throw error or not
}
```

I think it can happen that there are some cases when symbol is not exist in the Scope is not always the error case. And thought that checking whether it is an error or not is depends on the callers context.

image

For example, before we trying to update symbol table, we check whether symbol already exist. 
In this case when symbol already exist we should throw error and the case which symbol not exist in the Scope is ok condition. 

So I thought that the case which symbol is not existing in the Scope is not always the error case

junbeomlee (approved):

PR #241, opened 2019-02-03
test: bytecode Emerge

### LINKED ISSUE #221: [compiler] Implement TestBytecode_Emerge()
**Subject**
compiler

**Detail**
Modify `TestBytecode_Emit(t *testing.T)` to `TestBytecode_Emerge(t *testing.T)`!
Then, implement `TestBytecode_Emerge(t *testing.T)` test case in `translate/bytecode_test.go`

zeroFruit:
resolved: #221 

details:

1. test not overflowing digit for operands
2. test overflowing digit for operands

- [x] Test case

### DIFF
--- a/translate/bytecode_test.go
+++ b/translate/bytecode_test.go
@@ -16,9 +16,127 @@
 
 package translate_test
 
-import "testing"
+import (
+	"bytes"
+	"testing"
 
-// TODO: implement test cases :-)
-func TestBytecode_Emit(t *testing.T) {
+	"github.com/DE-labtory/koa/opcode"
+	"github.com/DE-labtory/koa/translate"
+)
 
+func TestBytecodeEmerge(t *testing.T) {
+	tests := []struct {
+		op       opcode.Type
+		operands [][]byte
+		result   translate.Bytecode
+	}{
+		{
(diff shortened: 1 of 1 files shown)


boohyunsik (commented):
Testing function names are different in each package.
How about talking about using same function name?

zeroFruit:
~~I would fix that 👍~~ 

done!

translate/bytecode_test.go
```diff
@@ -16,9 +16,127 @@
...
+				{0xff},
+				{0xff},
+			},
+			result: translate.Bytecode{
+				RawByte: []byte{byte(opcode.Mload), 0xff, 0xff},
+				AsmCode: []string{"Mload", "ff", "ff"},
+			},
+		},
+		// How to deal with byte overflow
+		{
```
hihiboss, line 67: Does `Loadfunc` need only 1 operand?

zeroFruit: Yes I intentionally put 2 operands to show that num of operands is not check here

If we add checking logic to this func, these tcs should be blocked

boohyunsik (approved):
Good

PR #248, opened 2019-02-04
Implements compile string

boohyunsik:
Implements compile string

- [x] Test case

### DIFF
--- a/translate/compiler.go
+++ b/translate/compiler.go
@@ -117,7 +117,7 @@ func compileStatement(s ast.Statement, bytecode *Bytecode, tracer MemTracer) err
 		return compileBlockStatement(statement, bytecode, tracer)
 
 	case *ast.ExpressionStatement:
-		return compileExpressionStatement(statement, bytecode)
+		return compileExpressionStatement(statement, bytecode, tracer)
 
 	case *ast.FunctionLiteral:
 		return compileFunctionLiteral(statement, bytecode)
@@ -143,7 +143,7 @@ func compileStatement(s ast.Statement, bytecode *Bytecode, tracer MemTracer) err
 // 	[value]
 //
 func compileAssignStatement(s *ast.AssignStatement, bytecode *Bytecode, memDefiner MemDefiner) error {
-	if err := compileExpression(s.Value, bytecode); err != nil {
+	if err := compileExpression(s.Value, bytecode, memDefiner); err != nil {
 		return err
 	}
 
@@ -184,8 +184,8 @@ func compileBlockStatement(s *ast.BlockStatement, bytecode *Bytecode, tracer Mem
 	return nil
 }
 
-func compileExpressionStatement(s *ast.ExpressionStatement, bytecode *Bytecode) error {
-	return compileExpression(s.Expr, bytecode)
+func compileExpressionStatement(s *ast.ExpressionStatement, bytecode *Bytecode, tracer MemTracer) error {
+	return compileExpression(s.Expr, bytecode, tracer)
 }
 
 // TODO: implement me w/ test cases :-)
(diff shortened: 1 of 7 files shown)


zeroFruit:
LGTM?! one feedback is that it would be better if there are some more (1 or 2?) tcs although it has similar concept

boohyunsik:
> LGTM?! one feedback is that it would be better if there are some more (1 or 2?) tcs although it has similar concept

Yeap i want to confirm that my implementation is right way :-)

junbeomlee:
mStore needs two operand such as offset and size of value!
and value should be stored in memory table!!

junbeomlee (changes requested):
mStore needs two operand such as `offset` and `size` of value!
and value should be stored in memory table!!

hihiboss:
When `MemoryTable` is merged to develop, add `Define()` to `compileStringLiteral()`.

hihiboss:
@junbeomlee @boohyunsik 

No, the value is not stored in memory table. In memory table, only the offset and the size of the value are stored. The value is stored in memory, using `Mstore`. `Mstore` needs the offset, size and value.

Suppose that `"abc"` is compiled. Since it is a string, you must use `Mstore`. `Mstore` needs three operands like the offset, size and value. First, The offset would be the memory counter of memory table. Second, the size would be `len("abc")`. Third, the value would be `"abc".`

**(Exactly, not `616263`(the bytes of "abc")` and `len([]byte{0x61, 0x62, 0x63})`, but `6162630000000000` and len([]byte{0x61, 0x62, 0x63, 0x00, 0x00, 0x00, 0x00, 0x00})`. Because we treat all data as 8 bytes.)**

So, the byte code would be
```asm
Push 6162630000000000    // "abc"
Push 0000000000000008    // size
Push 0000000000000000    // offset
Mstore
```
(if the current memory counter is 0.)

hihiboss:
Rebase and add a logic which pushes offset and size!

translate/compiler.go
```diff
@@ -303,13 +303,36 @@ func compileIntegerLiteral(e *ast.IntegerLiteral, bytecode *Bytecode) error {
...
-// TODO: implement me w/ test cases :-)
-func compileStringLiteral(e *ast.StringLiteral, bytecode *Bytecode) error {
+func compileStringLiteral(e *ast.StringLiteral, bytecode *Bytecode, memDefiner MemDefiner) error {
+	operand, err := encoding.EncodeOperand(e.Value)
+	if err != nil {
+		return err
+	}
+
+	for len(operand) >= 8 {
+		memEntry := memDefiner.Define(e.Value)
```
hihiboss: In `Define(id string)` , `id` means a key like variable name, not a value. When `int a` is saved in memory, compiling string `"a"` overwrites the memory table. Memory table just saves the location and size of value.
Also, The string literal doesn't have any meaning alone. The statement like assign statement has some meaning.

So, i think it should be
```go
size, err := encoding.EncodeOperand(len(operand))
if err != nil {
        return err
}

offset, err := encoding.EncodeOperand(memDefiner.MemoryCounter)
if err != nil {
        return err
}
```

If so, `Define()` is not used in here. Then it seems better that `memDefiner` is changed to `memEntryGetter` or `memTracer`.

translate/compiler_internal_test.go
```diff
@@ -891,10 +891,86 @@ func TestCompileIntegerLiteral_negative(t *testing.T) {
...
+					0x21, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x08, // Push 0x0000000000000000 (offset)},
+					0x23, // Memory store
+				},
+				AsmCode: []string{"Push", "7373737373737373",
+					"Push", "0000000000000008",
+					"Push", "0000000000000000",
+					"Push", "3131313131313100",
+					"Push", "0000000000000008",
+					"Push", "0000000000000008",
+					"Mstore"},
```
hihiboss, line 944: In this case, it just saves `3131313131313100`, not `7373737373737373`. 

Before executing `Mstore`, the stack would be
```asm
0000000000000008
0000000000000008
3131313131313100
0000000000000000
0000000000000008
7373737373737373
```
And, executing `Mstore`, it pops 3 data like `0000000000000008`, `0000000000000008`, `3131313131313100` and saves `3131313131313100` to memory. Then vm executes other opcode. 
Stack would be 
```asm
0000000000000000
0000000000000008
7373737373737373
```

So, I think `Mstore` is needed more. But, i know that we need to talk about longer string :)

translate/compiler.go
boohyunsik: Now memEntryGetter and memTracer both don't produce MemCounter because of abstraction.
How about use memTracer global var?

PR #249, opened 2019-02-04
feat: implement CallFunc arguments w/ tcs

### LINKED ISSUE #214: [vm] Implement arguments method with test case
**Subject**
vm.go

**Detail**
It is related to #201 

Example)
Pointer : 4bytes
Size : 4bytes
Value : dynamic
arguments(n) : if the number of arguments is 4, range of n is 0~3
cf.Args[n*4:n*4+4] : Pointer which point to value's size
after we know size, next to size is value.

zeroFruit:
resolved: #214 

details:

1. implements `arguments` logic w/ tcs

- [x] Test case

### DIFF
--- a/vm/vm.go
+++ b/vm/vm.go
@@ -23,6 +23,14 @@ import (
 	"github.com/DE-labtory/koa/opcode"
 )
 
+const (
+	// PTRSIZE is size of arguments pointer
+	PTRSIZE = 4
+
+	// SIZEPTRSIZE is size of arguments size pointer
+	SIZEPTRSIZE = 4
+)
+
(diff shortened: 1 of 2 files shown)


AgwaB:
It seems that the transfer of meaning is wrong!

I intented below feature and i also implemented `Encode` function like below!
```go
-----------------------------------------------------------------
 ptr1 | ptr2 | size1 | value1 | size2 | value2 
-----------------------------------------------------------------
```

AgwaB:
LGTM

boohyunsik (approved):
Good

PR #252, opened 2019-02-04
Feature/memory tracer

### LINKED ISSUE #232: [compiler] Remove Use() function in memory table.
**Subject**
compiler

**Detail**
Remove `Use()` function in `translate/memory_table.go`. `Use()` function was made for counting memory pointer in not assignment case. But, if using memory is not derived from the assignment, you should overwrite the memory! So, `Use()` function is no longer necessary.

hihiboss:
resolved: #232 

details:

1. Implemented Define() function in memory tracer.
2. Implemented test cases

- [x] Test case

==

Memory Tracer is used to memorize the memory location of the variable.
So, when you assign the value to variable using memory, you should use Define() function.

### DIFF
--- a/translate/tracer.go
+++ b/translate/tracer.go
@@ -16,6 +16,8 @@
 
 package translate
 
+import "github.com/pkg/errors"
+
 // MemoryTableEntry saves size and offset of the value which the variable has.
 type MemoryTableEntry struct {
 	Offset uint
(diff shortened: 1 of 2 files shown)


AgwaB:
I think we should check `id` duplication.
We didn't yet discuss about reallocation!

junbeomlee:
how about defining interface for memory table and add get function to get offset and size of the memory entry!

junbeomlee:
and it would be better to add test case checking last MemoryCounter value!

junbeomlee:
except that looks good!

hihiboss:
@leesd556 
Because parser checks duplication of the variables using scope, is `id` duplication not necessary?
In `memory table`, the function removing definition is not exist. So, I think that `Define()` should overwrite them.

translate/tracer.go
```diff
@@ -28,14 +42,66 @@ type MemoryTable struct {
 	MemoryCounter uint
 }
 
-// TODO: implement me w/ test cases :-)
+type MemoryTableDefiner interface {
+	Define(id string, value interface{}) (MemoryTableEntry, error)
```
zeroFruit: why separate `Definer` and `Getter`?

caller of `MemoryTable` separated?

hihiboss: Because of the go style interface which has only one method. I’ll combine them!

junbeomlee (approved):
Nice work!

PR #253, opened 2019-02-06
feat: add ReassignStatement to AST

zeroFruit:
details:

add `ReassignStatement` to AST

for example 

```
int foo = 1
foo = 2
```

- [x] Test case

### DIFF
--- a/ast/ast.go
+++ b/ast/ast.go
@@ -156,6 +156,22 @@ func (as *AssignStatement) String() string {
 	return out.String()
 }
 
+// ReassignStatement is used when we want re-assign value to variable
+type ReassignStatement struct {
+	Variable *Identifier
+	Value    Expression
+}
+
+func (rs *ReassignStatement) do() {}
+func (rs *ReassignStatement) String() string {
+	var out bytes.Buffer
+
+	out.WriteString(rs.Variable.String() + " = ")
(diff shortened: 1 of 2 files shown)


junbeomlee (approved):
LGTM

boohyunsik (commented):
Q) Can't we use assignStatement?

zeroFruit:
You mean can we reuse `AssignStatement`? yes! and if we use `AssignStatement` for re-assignment statement, compiler codes would be like this

```go
func compileStatement(s ast.Statement, bytecode *Bytecode) error {
	switch statement := s.(type) {
	case *ast.AssignStatement:
                if statement.Datastructure == nil {
                          return compileReassignStatement(statement, bytecode)
                } else {
                          return compileAssignStatement(statement, bytecode)
                }
		
```
vs if we create `ReassignStatment`
```go
func compileStatement(s ast.Statement, bytecode *Bytecode) error {
	switch statement := s.(type) {
	case *ast.AssignStatement:
                return compileAssignStatement(statement, bytecode)
        case *ast.ReassignStatement:
                return compileReassignStatement(statement, bytecode)
```

So I thought that it is better to create `ast.ReassignStatement` and for having clear meaning

junbeomlee:
i agree with @zeroFruit

PR #261, opened 2019-02-06
Add abi/type proto

### LINKED ISSUE #255: [abi] implement abi/Type.go
**Detail**
implement abi/Type.go

AgwaB:
resolved: #255

details:
Add abi/type proto

Now, we use jsut `int`, `bool`, `string`.
and `int` is basically `int64` because the maximum value of our stack is int64.

It is really basic Implementation what we need right now!
I considered we can add something new type(koa defined) later

- [ ] Test case


junbeomlee (approved):
LGTM

PR #262, opened 2019-02-06
Implements Arguments.Pack() func

### LINKED ISSUE #247: [vm] Implements Pack method with test case
**Subject**
abi/argument.go

**Detail**
Implemtns `Pack` method with test case
this is method of `Arguments` , not `Argument`.  
`Pack` will return Types of `Arguments` as string.

Ex)
If arguments has `int`, `bool`, `string` type, `Pack` will return  `int,bool,string`

This issue must precede #242 

**Reference**
write here

AgwaB:
resolved: #247 

details:
Implements Arguments.Pack() func

- [ ] Test case

### DIFF
--- a/abi/argument.go
+++ b/abi/argument.go
@@ -16,14 +16,49 @@
 
 package abi
 
+import (
+	"encoding/json"
+	"strings"
+)
+
 type Argument struct {
 	Name string
-	Type string
+	Type Type
 }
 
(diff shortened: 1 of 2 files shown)


junbeomlee (approved):
nice work!

zeroFruit (approved):

PR #264, opened 2019-02-06
Implements checking scope feature in parseCallArgs

### LINKED ISSUE #258: [parser] check symbol table when parsing function parameters
**Detail**
```go
func parseFunctionParameter(buf TokenBuffer) (*ast.ParameterLiteral, error) {
	token := buf.Read()
	if token.Type != Ident {
		return nil, ExpectError{
			token,
			Ident,
		}
	}

	// TODO: check whether variable name exist,
	// TODO: but do not update symbol table in this case
	// TODO: if not, return error
```

boohyunsik:
resolved: #258 

Implements checking scope feature in parseCallArgs

- [x] Test case

### DIFF
--- a/parse/parser.go
+++ b/parse/parser.go
@@ -663,23 +663,20 @@ func parseFunctionParameter(buf TokenBuffer) (*ast.ParameterLiteral, error) {
 		}
 	}
 
-	// TODO: check whether variable name exist,
-	// TODO: but do not update symbol table in this case
-	// TODO: if not, return error
-
 	ident := &ast.ParameterLiteral{
 		Identifier: &ast.Identifier{Value: token.Val},
 	}
 
-	token = buf.Read()
-	ds, ok := datastructureMap[token.Type]
+	dsToken := buf.Read()
+	ds, ok := datastructureMap[dsToken.Type]
 	if !ok {
 		return nil, Error{
-			token,
+			dsToken,
 			"Function parameter type missed",
 		}
 	}
 	ident.Type = ds
+	updateScopeSymbol(token, dsToken)
 
 	return ident, nil
 }
@@ -819,10 +816,12 @@ func parseCallArguments(buf TokenBuffer) ([]ast.Expression, error) {
 	if err != nil {
 		return nil, err
 	}
+
 	args = append(args, exp)
 
 	for curTokenIs(buf, Comma) {
 		buf.Read()
+
 		exp, err := parseExpression(buf, LOWEST)
 		if err != nil {
 			return nil, err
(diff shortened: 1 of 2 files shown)


zeroFruit:
plz refer to comment on issue #258

parse/parser.go
```diff
@@ -637,9 +637,10 @@ func parseFunctionParameter(buf TokenBuffer) (*ast.ParameterLiteral, error) {
 		}
 	}
 
-	// TODO: check whether variable name exist,
-	// TODO: but do not update symbol table in this case
-	// TODO: if not, return error
+	exist := scope.Get(token.Val)
```
zeroFruit: should use `updateScopeSymbol` so after #265 merged we can fix this issue

zeroFruit:
Logic is good but we need to fix tcs.

suggestion:

#### `TestParseFunctionParameter`

1. test after parsing is done, scope successfully saves symbols. You can find similar example in `TestUpdateScopeSymbol`

2. add test case when function parameters have same name: `func foo(a int, a string) {`

#### `TestParseCallArguments`

1. add returning error test case when scope has no symbol

parse/parser_internal_test.go
```diff
@@ -990,6 +990,7 @@ func TestParseFunctionParameter(t *testing.T) {
 	initParseFnMap()
 	tests := []struct {
 		buf         TokenBuffer
+		setupScope  setupScopeFn
 		expected    []string
 		expectedErr error
```
zeroFruit, line 1036: I think we almost done, and last comment is plz add chkScopeFn to `TestParseFunctionParameter` to check whether symbols are successfully saved (rebase first! and you will find similar pattern in `ParseIfStatement`)

boohyunsik: chkScopeFn checks only current scope and parent scope.
parseBlockStatement does not have entering and leaving scope feature currently. (It is being reviewed.)
If program <b>leaves current scope</b>, chkScopeFn will return false always..
For example,
```go
if (true) {
   int a = 1
}
```
We want to check the value "a" is in right scope, but after parsing above, <b>current scope</b> is outer scope. so chkScopeFn will be failed.(Our scope checking logic goes to the outer scope, as you know.)

zeroFruit: #290 pr created, after merge we can fix this issue!

boohyunsik: We have another problem!
Now, we can go to inner scope. It is same as we can search root to leaf. (Using dfs, we can top-down process.)
But how about this example?

```go
if (a == 5) {
    int b = 3;
} else {
    int b = 4;
}
```

If we want to check "b" in <b>else block</b>, but parser may check <b>if block first.</b>
I think it is incorrect.

zeroFruit (approved):

Issue #267, opened 2019-02-07
[parser] Need to refactor parser_test


boohyunsik:
We need to refactor parser_test because of adding scope.
Many errors occur due to scope!

zeroFruit:
temporarily going to comment out parser external test!

junbeomlee:
i agree!

PR #268, opened 2019-02-07
feat: support symbol checking when parsing assign stmt

### LINKED ISSUE #259: [parser] check symbol table when parsing assignment statement
**Detail**
```go
// parseAssignStatement parse assign statements which assign values
// to its identifier. e.g. int a = 1
func parseAssignStatement(buf TokenBuffer) (*ast.AssignStatement, error) {
	stmt := &ast.AssignStatement{}

	token := buf.Read()

	ds := datastructureMap[token.Type]
	stmt.Type = ds

	token = buf.Read()
	if token.Type != Ident {
		return nil, ExpectError{
			token,
			Ident,
		}
	}

	// TODO: check whether variable name exist
	// TODO: if not, add symbol to the scope
```

zeroFruit:
resolved: #259 

1. type check when parsing assign statement
2. **commented out parser integral tests (#267), when finishing scope feature, this test should be changed**

- [x] Test case

### DIFF
--- a/parse/parser.go
+++ b/parse/parser.go
@@ -698,12 +698,10 @@ func parseGroupedExpression(buf TokenBuffer) (ast.Expression, error) {
 func parseAssignStatement(buf TokenBuffer) (*ast.AssignStatement, error) {
 	stmt := &ast.AssignStatement{}
 
+	dsToken := buf.Read()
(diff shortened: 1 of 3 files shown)


boohyunsik (approved):
LGTM

Issue #270, opened 2019-02-07
[parser] implements parseReassignStatement w/ tcs


zeroFruit:
**Detail**

implements parseReassignStatement w/ tcs

PR #273, opened 2019-02-08
Implements scope in block statement

### LINKED ISSUE #269: [parser] apply scope to block statements
**Detail**

```go
func parseBlockStatement(buf TokenBuffer) (*ast.BlockStatement, error) {
	if err := expectNext(buf, Lbrace); err != nil {
		return nil, err
	}

	// TODO: enter scope


	for curToken.Type != Rbrace && curToken.Type != Eof {
		...
	}

	if curTokenIs(buf, Rbrace) {
		buf.Read()
	}

	// TODO: leave scope

	return block, nil
}
```

boohyunsik:
resolved: #269, #270

1. Implements parseReassignStatmenet and tcs
2. Implements scoping in block statement

- [x] Test case

### DIFF
--- a/parse/parser.go
+++ b/parse/parser.go
@@ -369,8 +369,8 @@ func parseStatement(buf TokenBuffer) (ast.Statement, error) {
 		return parseIfStatement(buf)
 	case Return:
 		return parseReturnStatement(buf)
-	case Lbrace:
-		return parseBlockStatement(buf)
+	case Ident:
+		return parseReassignStatement(buf)
 	default:
 		return parseExpressionStatement(buf)
 	}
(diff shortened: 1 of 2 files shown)


parse/parser.go
```diff
@@ -346,6 +356,10 @@ func parseStatement(buf TokenBuffer) (ast.Statement, error) {
 		return parseIfStatement(buf)
 	case Return:
 		return parseReturnStatement(buf)
+	case Lbrace:
```
zeroFruit: I think we don't support 

```
func foo() {
  {
      int a = 1
  }
}
```
such statement. So we don't need `Lbrace` case now

boohyunsik: We may implement this after scoping problem solved.

zeroFruit (approved):
👍 👍

PR #274, opened 2019-02-08
add compile infix function

hea9549:
SOMEBODY GIVE ME T/C HAHAHAHA


hihiboss:
`land` or `lor` are also compiled with `eq`??

boohyunsik (changes requested):
ADD TC HAHAHA

boohyunsik (approved):

PR #287, opened 2019-02-10
Add limitation to parsing prefix function

boohyunsik:
Add limitation to parsing prefix function
like below
```go
!a
-a   // a is string
!3
```

- [x] Test case


zeroFruit (approved):
Nice work!

PR #290, opened 2019-02-11
feat: add inner to scope

zeroFruit:
details:

1. add `inner` field to `Scope`, to access inner scope. This makes test easy and possible
2. fix `enterScope` logic

- [x] Test case


boohyunsik (approved):
Good

PR #294, opened 2019-02-11
doc: compiler.

hihiboss:
~resolved: #[issue number]~

details:

Documentation the compiler.

~- [ ] Test case~


joomanzi:
how about adding assignment operator such as += , -=, ...
i researched other language spec, some has section for assignment operator and some has not.
so i think it depends on our hand. we can count on this

hihiboss:
@joomanzi 
Good idea! But, I think they are not matched to our concept. Our concept is Simple!
However, that is just my personal opinion. I wonder the other's thought. 

Even if it is decided to add those operators, this feature should be implemented in milestone 0.2.0 or 0.3.0 :)
(Because lexer, parser, compiler, vm.. everyone has to add that feature.)

joomanzi:
lexer already got those feature!

hihiboss:
@joomanzi
Oh, really?! Nice work! I'll add that feature later!

junbeomlee:
Nice work!

junbeomlee (approved):
LGTM

PR #297, opened 2019-02-11
Add test case when the integer value is negative

### LINKED ISSUE #286: [compiler] Add test case when the integer value is negative
**Subject**
compiler

**Detail**
Add test case when the integer value is negative in `TestCompileIntegerLiteral()`

AgwaB:
resolved: #286

Add test case when the integer value is negative

### DIFF
--- a/translate/compiler_internal_test.go
+++ b/translate/compiler_internal_test.go
@@ -787,6 +787,31 @@ func TestCompileIntegerLiteral(t *testing.T) {
 	runExpressionCompileTests(t, tests)
 }
 
+func TestCompileIntegerLiteral_negative(t *testing.T) {
+	tests := []expressionCompileTestCase{
+		{
+			expression: &ast.IntegerLiteral{
+				Value: -10,
+			},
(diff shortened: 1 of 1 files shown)


translate/compiler_internal_test.go
```diff
@@ -787,6 +787,31 @@ func TestCompileIntegerLiteral(t *testing.T) {
...
+		},
+		{
+			expression: &ast.IntegerLiteral{
+				Value: -20,
+			},
+			expected: Bytecode{
+				RawByte: []byte{0x21, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xEC},
+				AsmCode: []string{"Push", "ffffffffffffffec"},
+			},
+		},
```
boohyunsik, line 809: If -0? enable?

AgwaB: Yes. Recognize as 0.

hihiboss (approved):
TYVM!

PR #305, opened 2019-02-13
Feature/function jumper

### LINKED ISSUE #279: [compiler] Implement generateFuncJumper()
**Subject**
compiler

**Detail**
Implement `generateFuncJumper()` function in `translate/compiler.go`.

hihiboss:
resolved: #279 

details:

1. Define `FuncMap` which has the location(program counter) of each function.
2. Implemented Function Jumper!

**Compile Flow**
1. compile each function.
2. expect function jumper size (because the size depends on the number of functions).
3. generate function jumper.
4. combine the function jumper and the byte code (which is compiled for function logic).

**Function Jumper Logic**
1. load call function selector (`LoadFunc`)
2. duplicate call function selector.
3. push the function selector of function being compiled.
4. compare call function selector with the selector being compiled.
5. push the location(program counter) to jump.
6. repeat 2~5 for each function.
7. add logic to revert (nothing corresponding to call function selector).

reference: https://blog.zeppelin.solutions/deconstructing-a-solidity-contract-part-iii-the-function-selector-6a9b6886ea49

*Test case not yet. Plz wait :)*

- [ ] Test case

### DIFF
--- a/translate/compiler.go
+++ b/translate/compiler.go
@@ -28,9 +28,9 @@ import (
 
 type FuncMap map[string]int
 
-func (m FuncMap) Declare(f ast.FunctionLiteral, b Bytecode) {
-	funcSel := abi.Selector(f.Signature())
-	m[string(funcSel)] = len(b.RawByte)
+func (m FuncMap) Declare(signature string, b Bytecode) {
+	funcSel := abi.Selector(signature)
+	m[string(funcSel)] = len(b.AsmCode)
 }
 
 // TODO: implement me w/ test cases :-)
@@ -42,19 +42,14 @@ func CompileContract(c ast.Contract) (Bytecode, error) {
 		AsmCode: make([]string, 0),
 	}
 
-	memTracer := NewMemEntryTable()
-
 	funcMap := FuncMap{}
-
-	expectedFuncJmpr, err := expectFuncJmpr(c)
-	if err != nil {
+	if err := expectFuncJmpr(c, bytecode, funcMap); err != nil {
 		return *bytecode, err
 	}
 
-	bytecode.Append(expectedFuncJmpr)
-
+	memTracer := NewMemEntryTable()
 	for _, f := range c.Functions {
-		funcMap.Declare(*f, *bytecode)
+		funcMap.Declare(f.Signature(), *bytecode)
 
 		if err := compileFunction(*f, bytecode, memTracer); err != nil {
 			return *bytecode, err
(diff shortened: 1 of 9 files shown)


translate/compiler.go
```diff
@@ -26,6 +26,13 @@ import (
 	"github.com/DE-labtory/koa/opcode"
 )
 
+type FuncMap map[string]int
+
+func (m FuncMap) Declare(f ast.FunctionLiteral, b Bytecode) {
+	funcSel := abi.Selector(f.Signature())
```
AgwaB: How about `funcSig` ?

translate/compiler_test.go
```diff
@@ -16,9 +16,82 @@
...
-import "testing"
+import (
+	"testing"
+
+	"github.com/DE-labtory/koa/abi"
+	"github.com/DE-labtory/koa/ast"
+	"github.com/DE-labtory/koa/translate"
+)
+
+func TestFuncMap_Declare(t *testing.T) {
```
AgwaB, line 27: plz add testcase when there are two more functions.

translate/compiler.go
```diff
@@ -37,19 +44,102 @@ func CompileContract(c ast.Contract) (Bytecode, error) {
...
+	// Add function jumper to bytecode
+	// [bytecode] => [function jumper][bytecode]
+	funcJpr.Append(bytecode)
+	bytecode = funcJpr
+
+	return nil
+}
+
+// TODO: implement me w/ test cases :-)
+func compileFunctionLiteral(s *ast.FunctionLiteral, bytecode *Bytecode, funcMap FuncMap) error {
```
zeroFruit: `ast.FunctionLiteral` contains not only `Name`, `Parameters` but also `Body`. so I thought that `compileFunctonLiteral` was compiling entire function. 

But this is about compiling one of function jumper switch cases. So little bit confusing

translate/compiler.go
```diff
@@ -37,19 +44,102 @@ func CompileContract(c ast.Contract) (Bytecode, error) {
...
 		return *bytecode, err
 	}
 
 	return *bytecode, nil
 }
 
+// Expects a size of the function jumper and updates the function map.
+func expectFuncJumper(c ast.Contract, funcMap FuncMap) error {
+	expectedJpr := NewBytecode()
+	if err := generateFuncJumper(c, expectedJpr, funcMap); err != nil {
```
zeroFruit: `generateFuncJumper` here is for calculating function jumper bytes size? 

If so, in my intuition we can calculate the fn jumper size right away(because fn sel is 4 bytes), not after generating function jumper.

I comment this because I'm a little bit consufed with `generateFuncJumper`

translate/compiler.go
hihiboss: Instead of `funcSel`?

translate/compiler_test.go
hihiboss: For `Declare`? This function handles just one `FunctionLiteral`. And, test `bar` and `sam` declares when there are two or more functions.

translate/compiler.go
hihiboss: `Body` in `ast.FunctionLiteral` would be compiled in `compileFunction()`, not `compileFunctionLiteral()`. `compileFunctionLiteral()` just get the function selector and add a logic to find the function location called. I think `compileFunctionLiteral()` name is so confusing. How about `compileFuncSelector()`? If you have any idea, please recommend it!

translate/compiler.go
hihiboss: Yes, you're right. I had concern to them, but i couldn't any solution.

Function Jumper logic is the following.

First, compile functions. This means compiling a logic for each function and find the location of that function in byte code. 
Second, expect function jumper.
...

Suppose that the byte code is `00616263` and the start of function to compile (`func foo`) is `63`. Then, in that byte code, `func foo` location is 3. (0: `00`, 1: `61`, 2: `62`, 3: `63`) So in `FuncMap`, `FuncMap["foo"]` is `3`. But if function jumper is added in front of the byte code, the function location is changed. For example, function jumper is `112233`, the byte code should be `11223300616263` and `FuncMap["foo"]` should be `6`. So, I have to update `FuncMap` after adding the function jumper to byte code. In this time, a problem occur. To update `FuncMap`, i must know a size of function jumper. But, to know the function jumper, I generate a function jumper. Then, you might think that 'Compile functions, generate function jumper, update `FuncMap`, then combine function jumper and byte code!'. But, After updating `FuncMap`, I regenerate a function jumper. Because, as changing `FuncMap`, the locations of functions are also changed. 

So, compile functions, generate function jumper to expect the size of function jumper, update `FuncMap`, generate function jumper with updated `FuncMap`, and combine.

If you have a good idea, please tell me!! :)

translate/compiler.go
boohyunsik: I think compileFuncSelector() is good.

translate/compiler_internal_test.go
```diff
@@ -30,11 +30,23 @@ type expressionCompileTestCase struct {
 	expected   Bytecode
 }
 
-// TODO: implement test cases :-)
+func TestExpectFuncJumper(t *testing.T) {
```
boohyunsik: Recommendation) Write tc first. It may be helpful. :-)

translate/compiler.go
zeroFruit: OK I understand your intention about this function. 

If my understanding is right, this function's job is 

(1) create function jumper (just for calculating its size) then  
(2) relocate function map based on its size

So `relocateFuncMapEntry` would be better. (just suggestion :) )

zeroFruit: And how about receiving function jumper size as arguments?
```go
func expectFuncJumper(funcMap FuncMap, fnJmprSize Size) error { ... }
```

I thought that we don't need to create function jumper, just receive function jumper size from outside is enough (if we change like this, `ast.Contract` also not needed)

And if we decided `expectFuncJumper` receive function jumper size from arguments, I think we can change the logic in **line 57 ~ 61**. 

```go
if err := expectFuncJumper(c, funcMap); err != nil {
	return *bytecode, err
}

if err := generateFuncJumper(c, bytecode, funcMap); err != nil {
	return *bytecode, err
}
```
to
```go
if err := generateFuncJumper(c, bytecode, funcMap); err != nil {
	return *bytecode, err
}
if err := expectFuncJumper(funcMap, len(bytecode)); err != nil {
	return *bytecode, err
}
```

Issue #311, opened 2019-02-14
[parser] parseReassignStatement should be check in default cases


zeroFruit:
**Where**
parser

**Bug detail**
```
add(1, 2)

a = 1
```

I think both should be handled in default case in `parseStatement`

**Fix plan**

lookup one more token then decide whether this is reassign or callexpression

zeroFruit:
@zeroFruit

PR #315, opened 2019-02-15
bug: parsingReassignStatement, parsing CallExpression

### LINKED ISSUE #314: [parser] not parsing when empty parameter function comes in as function arguments
**Where**
parser

**Bug detail**
```
foo(add(), 1)
```
parsing results

```
function foo( function add(  ) )
```

arguments after comma not parsed

zeroFruit:
resolved: #314, #311 

details:

1. not parsing when empty parameter function comes in as function arguments
2. parseReassignStatement should be check in default cases

- [x] Test case


junbeomlee (approved):
LGTM

PR #316, opened 2019-02-16
bug: enter/leave scope when start parsing function literal

### LINKED ISSUE #309: [parser] when parsing contract block, should enter/leave scope
**Where**
parser

**Bug detail**
```
func add(int a, int b) {}

func foo(int a, int c) {}
```
is not allowed, because `a` is added scope when parsing `add` and when parsing `foo` because `a` is already added to scope this throw an error.

zeroFruit:
resolved: #309

details:

enter/leave scope when start parsing function literal, as a result we can now parsing like this situation:

```
contract {
    func foo(a int) {}
    func bar(a int) {} // now we can use `a` in the parameter
}
```

- [x] Test case


junbeomlee (approved):
LGTM!

PR #317, opened 2019-02-16
bug: return void statement

### LINKED ISSUE #307: [parser] bug in returning void
**Where**
parser

**Bug detail**
currently our parser not allow returning void such as
```
func foo() {
    return
}
```

zeroFruit:
resolved: #307 

details:

1. parse return void statement
2. fix `StringLiteral` `toString` function logic

- [x] Test case


boohyunsik (approved):
LGTM

PR #330, opened 2019-02-17
fix: pop the expression in the stack when compiling expressionStatement

### LINKED ISSUE #327: [compiler] Pop the expression after compiling ONE statement
**Subject**
compiler

**Detail**
After compiling a statement, the expression which came out in `compileStatement()` remains in the stack. So, If `compileStatemnt()` has an expression result, you should pop that expression.

I think it needs in `compileExpressionStatement()`.

hihiboss:
resolved: #327

details:

Fix bug that when compiling expressionStatement, the expression remains in the stack.

- [x] Test case


zeroFruit (approved):

zeroFruit:
rsv the conflict

PR #331, opened 2019-02-17
Implements compileFunction() and tcs

boohyunsik:
Implements compileFunction() and tcs

- [x] Test case


hihiboss (commented):
Could you add a case which use the parameter? :)

Ex) 
func foo(int a) { return a}

hihiboss (dismissed):
Nice work!

PR #333, opened 2019-02-18
feat: fix if-statement compile bug

### LINKED ISSUE #326: [compiler] If and IfElse statement compile bug
**Where**
compiler

**Bug detail**
When compiling nested if or if-else condition statement, pc to jump in inner condition statement could be changed.

junbeomlee:
resolved: #326 

details:

fix bug in if-statement compile

- [x] Test case


hihiboss (approved):
LGTM!!!

PR #337, opened 2019-02-18
Implements integration test cases

boohyunsik:
integrate tcs using .koa files

### DIFF
--- a/koa_test.go
+++ b/koa_test.go
@@ -1,0 +1,108 @@
+/*
+ * Copyright 2018 De-labtory
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


koa_test.go
```diff
@@ -0,0 +1,372 @@
...
+
+	for i, test := range tests {
+		str, err := readFile(test.fileName)
+		if err != nil {
+			continue
+		}
+
+		asm, err := Compile(str)
+
+		// TODO : after implements all functions in compile
```
zeroFruit, line 363: What is todo?

boohyunsik: This is just TODO..because we currently are not implementing compile function and call.
So I want to mark TODO for making other cases after making those functions.

zeroFruit: ok!

zeroFruit (approved):

PR #345, opened 2019-02-21
Feature/vm/execute

AgwaB:
details:

@zeroFruit 

If we have this function,
```go
func A(a int, b int){
}
```
we can get function selector by `A(int,int)` 

but this method make as A(parameter : identifier : ....  maybe?

I think below method has matter so  plz modify!

```go
func (p *ParameterLiteral) String() string {
	return fmt.Sprintf("Parameter : (Identifier: %s, Type: %s)", p.Identifier.String(), p.Type.String())
}
```
- [x] Test case

### DIFF
--- a/vm/vm_test.go
+++ b/vm/vm_test.go
@@ -209,7 +209,7 @@ func TestExecute(t *testing.T) {
 			t.Error(err)
 		}
 
-		result := stack.pop()
+		result := stack.Pop()
 		if result != test.result {
 			t.Errorf("Invalid result - expected = %d, got = %d", test.result, result)
 		}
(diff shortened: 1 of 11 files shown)


vm/vm_test.go
```diff
@@ -1043,7 +1043,7 @@ func TestSWAP(t *testing.T) {
 	}
 
 	if len(stack.items) != 2 {
-		t.Errorf("Invalid stack size - expected=%d, got =%d", len(testExpected), stack.len())
+		t.Errorf("Invalid stack size - expected=%d, got =%d", len(testExpected), stack.Len())
 	}
 
 	for i, item := range stack.items {
```
boohyunsik, line 1049: LGTM but there is a miss typo in commit message :-)

junbeomlee (approved):
LGTM

PR #348, opened 2019-02-22
fix: add checking void type logic to abi

zeroFruit:
* fix: add checking void type logic to abi

- [x] Test case


hihiboss (approved):
LGTM

PR #354, opened 2019-02-23
fix: returns an error when handling undefined operator

### LINKED ISSUE #302: [compiler] Fix the bug which returns an error when handling undefined operator
**Where**
compiler

**Bug detail**
In `compileInfixExpression()`, there is a bug that returns an error when handling undefined operator.

**Fix plan**
Add error returning logic in default case and implement test case!

hihiboss:
resolved: #302 

details:

Fix the bug.

- [x] Test case


junbeomlee (approved):
LGTM

PR #358, opened 2019-02-26
Implements skel functions for scope refactoring

boohyunsik:
[WIP] Refactoring scope


symbol/scope.go
```diff
@@ -17,92 +17,168 @@
...
-func NewScope() *Scope {
-	s := make(map[string]Symbol)
-	return &Scope{
-		store: s,
-		inner: make([]*Scope, 0),
-		outer: nil,
-	}
+// Define scope errors
+type DupError struct {
+	Identifier ast.Identifier
```
zeroFruit: I think it is better `DupError` has `Symbol` not `Identifier`

symbol/scope.go
```diff
@@ -17,92 +17,168 @@
...
-	s.store[name] = val
-	return val
-}
+func ScopingFunction(f *ast.FunctionLiteral, s *Scope) error {
+	name := f.Name.Name
+	s.store[name] = &Function{Name: name}
+	s = NewEnclosedScope(s)
+
+	if err := ScopingFunctionParameter(f.Parameters, s); err != nil {
+		return nil
```
zeroFruit: return err

symbol/scope.go
```diff
@@ -17,92 +17,168 @@
...
+		return err
+	}
+
+	s = s.parent
+	return nil
 }
 
-func (s *Scope) GetOuter() *Scope {
-	return s.outer
+func ScopingFunctionParameter(f []*ast.ParameterLiteral, s *Scope) error {
```
zeroFruit: why `f` ? could have better name

symbol/scope.go
```diff
@@ -17,92 +17,168 @@
...
+	for _, p := range f {
+		switch p.Type {
+		case ast.IntType:
+			{
```
zeroFruit: remove {} plz..

symbol/scope.go
```diff
@@ -17,92 +17,168 @@
...
+		err = ScopingAssignStatement(implStmt, scope)
+	case *ast.BlockStatement:
+		err = ScopingBlockStatement(implStmt, scope)
+	case *ast.ExpressionStatement:
+		// TODO
+	case *ast.IfStatement:
+		err = ScopingIfStatement(implStmt, scope)
+	}
+
+	if err != nil {
```
zeroFruit: I think this if statement don't need, just `return err` enough

symbol/scope.go
```diff
@@ -17,92 +17,168 @@
...
-	inner []*Scope
-	outer *Scope
+	child  []*Scope
+	parent *Scope
+}
+
+func NewScope() *Scope {
+	return &Scope{
+		store:  make(map[string]Symbol),
+		parent: nil,
```
zeroFruit: don't need this line

zeroFruit (commented):
1. How about changing function from ScopingXXX to ResolveXXX, later these functions should do type checking.
2. How about changing Symbol to Object?

symbol/scope.go
boohyunsik: These are all WIP..:-)

symbol/resolver.go
```diff
@@ -17,41 +17,318 @@
...
+// Define scope structure
+type Scope struct {
+	store map[string]Object
+
+	child  []*Scope
+	parent *Scope
+}
+
+func (s *Scope) Get(name string) Object {
+	scope := s
```
hihiboss, line 84: Why use `scope` variable instead of `s`??

hihiboss: If you want to clone `Scope`, `scope := *s` or `func (s Scope) Get(name string) Object {}` might be right

hihiboss: Or `func (s Scope) Get(name string) Object {}`

symbol/resolver.go
```diff
@@ -17,41 +17,318 @@
...
+	obj := &Function{Name: name}
+
+	r.scope.store[name] = obj
+	r.fns[f] = FunctionObject
+	r.defs[f.Name] = obj
+
+	r.scope = NewEnclosedScope(r.scope)
+
+	if err := ResolveFunctionParameter(f.Parameters, r); err != nil {
+		return nil
```
hihiboss, line 169: Why return `nil`?

symbol/resolver.go
```diff
@@ -17,41 +17,318 @@
...
+	for _, p := range p {
+		if obj, ok := r.scope.store[p.Identifier.Name]; ok {
+			return DupError{
+				object: obj,
+			}
+		}
+
+		var obj Object
+		var t ObjectType
+		switch p.Type {
```
hihiboss, line 190: (refactor)

How about define `func resolveParamObj` like
```go
func resolveParamObj(p ast.ParameterLiteral) (Object, ObjectType) {
    switch p.Type {
        case ast.IntType:
            return &Integer{Name: p.Identifier}, IntegerObject
    ...
    }
```

And, 

in `func ResolveFunctionParameter()`
```go
for _, p := range p {
	if obj, ok := r.scope.store[p.Identifier.Name]; ok {
		return DupError{
			object: obj,
		}
	}
	
	obj, t := resolveParamObj(p)
	r.scope.store[p.Identifier.Name] = obj
	r.types[p] = t
	r.defs[p.Identifier] = obj
}
return nil

File: symbol/resolver.go [hunk 10]
```diff
@@ -17,41 +17,318 @@
...
+	if err := ResolveFunctionBody(f.Body, r); err != nil {
+		return err
+	}
+
+	r.scope = r.scope.parent
 	return nil
 }
+
+func ResolveFunctionParameter(p []*ast.ParameterLiteral, r *Resolver) error {
+	for _, p := range p {
```
hihiboss, line 181: Variable name is duplicated!
Although there is not any errors, i think it should be changed for readability.
ex) `for _, param := params`

symbol/resolver.go
```diff
@@ -17,41 +17,318 @@
...
+func ResolveFunctionBody(b *ast.BlockStatement, r *Resolver) error {
+	if err := ResolveBlockStatement(b, r); err != nil {
+		return err
+	}
+	return nil
+}
+
+func ResolveStatement(stmt ast.Statement, r *Resolver) error {
+	var err error
+	switch implStmt := stmt.(type) {
```
hihiboss, line 217:
```go
case *ast.AssignStatement:
		return ResolveAssignStatement(implStmt, r)
```
is better!

